PR #7, opened 2023-11-22
Index fix

lanserge:
In the LibertyReader the index of the current_waveforms was calculated as slew_index * slew_axis->size() + cap_index, but should be slew_index * cap_axis->size() + cap_index (same as in the OutputWaveforms::currentWaveform function).

This commit fix the occasional memory corruption issues caused by above index calculated wrong and been outside allocated array.

### DIFF
--- a/liberty/LibertyReader.cc
+++ b/liberty/LibertyReader.cc
@@ -2524,7 +2524,7 @@ LibertyReader::endOutputCurrentRiseFall(LibertyGroup *group)
     slew_axis->findAxisIndex(waveform->slew(), slew_index, slew_exists);
     cap_axis->findAxisIndex(waveform->cap(), cap_index, cap_exists);
     if (slew_exists && cap_exists) {
-      size_t index = slew_index * slew_axis->size() + cap_index;
+      size_t index = slew_index * cap_axis->size() + cap_index;
       current_waveforms[index] = waveform->stealCurrents();
       (*ref_times)[slew_index] = waveform->referenceTime();
     }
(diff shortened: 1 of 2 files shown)


lanserge:
Closed this as it somehow contains not needed merge request that was done in opposite direction.
Will reopen this pull request with only required commits.

search/Sta.cc
```diff
@@ -366,7 +366,8 @@ Sta::updateComponentsState()
 void
 Sta::makeReport()
 {
-  report_ = new ReportTcl();
```
parallaxsw, line 369: // Default constructors that are called by makeComponents in the Sta
 The intended way to handle this is with a derived class like the comment in Sta.hh indicates
 // constructor.  These can be redefined by a derived class to
  // specialize the sta components.
  virtual void makeReport();

liberty/LibertyReader.cc
```diff
@@ -2524,7 +2524,7 @@ LibertyReader::endOutputCurrentRiseFall(LibertyGroup *group)
     slew_axis->findAxisIndex(waveform->slew(), slew_index, slew_exists);
     cap_axis->findAxisIndex(waveform->cap(), cap_index, cap_exists);
     if (slew_exists && cap_exists) {
-      size_t index = slew_index * slew_axis->size() + cap_index;
```
parallaxsw, line 2527: Thanks, this has been updated in repo

search/Sta.cc
lanserge: Ah, ok, make sense, thank you

PR #8, opened 2023-11-23
Fixed memory corruption due to wrong calculation of the index

lanserge:
In the LibertyReader the index of the current_waveforms was calculated as slew_index * slew_axis->size() + cap_index, but should be slew_index * cap_axis->size() + cap_index (same as in the OutputWaveforms::currentWaveform function).

This commit fix the occasional memory corruption issues caused by above index calculated wrong and been outside allocated array.
      in the OutputWaveforms table while reading liberty file


lanserge:
Closing it as I can see the fix was added.

PR #11, opened 2024-03-05
Verilog attribute parsing OpenSTA

QuantamHD:
Allows STA to read, parse and write verilog attributes, example below.

```verilog
/* Generated by Yosys 0.8+ */

(* hdlname = "\\counter" *)
(* top =  1  *)
(* src = "synthesis/tests/counter.v:16.1-32.10" *)
module counter(clk, reset, out);
  wire _0000_;
  wire _0001_;
```


examples/attribute_parsing_test.tcl
jjcherry56, line 1: It is pointless to put "test" in a test file name. They are all tests.

examples/attribute_parsing_test.v
jjcherry56, line 1: I suggest constructing a test case that is far smaller than 5000 lines that
targets the various parsing scenarios.

include/sta/ConcreteLibrary.hh
```diff
@@ -38,6 +38,7 @@ class LibertyCell;
 class LibertyPort;
 
 typedef Map<const char*, ConcreteCell*, CharPtrLess> ConcreteCellMap;
+typedef Map<std::string, const char*> AttributeMap;
```
jjcherry56, line 41: why not string -> string?
20 years ago std::string was pretty worthless, but I am trying to make more use of it these days

network/ConcreteLibrary.cc
```diff
@@ -268,6 +269,18 @@ ConcreteCell::setIsLeaf(bool is_leaf)
...
 }
 
+void
+ConcreteCell::setAttribute(const char* key, const char* value)
+{
+  attribute_map_.insert(key, stringCopy(value));
+}
+
+const char *
+ConcreteCell::getAttribute(const char* key) const
```
jjcherry56, line 279: trailing space

include/sta/ConcreteLibrary.hh
```diff
@@ -105,6 +106,8 @@ public:
   ConcreteCellPortBitIterator *portBitIterator() const;
   bool isLeaf() const { return is_leaf_; }
   void setIsLeaf(bool is_leaf);
+  void setAttribute(const char *key, const char *value);
```
jjcherry56, line 109: The formatting convention is for each function argument to be on a separate line.
This comment applies to most of the multi-arg functions defined in the PR.

include/sta/ConcreteNetwork.hh
```diff
@@ -206,6 +212,9 @@ public:
   Pin *connect(Instance *inst,
                LibertyPort *port,
                Net *net) override;
+  void setAttribute(Instance *inst,
+		       const char *key,
```
jjcherry56, line 216: indentation messed up here and many other places.
Probably an issue with tabs. I set emacs to not use tabs so the tab
setting does not matter. I tried for many many hours to configure clang-format
conform to the indentation standards and could not get it to work (even though
twenty years ago I chose the formating based on google's standard).

include/sta/ConcreteLibrary.hh
QuantamHD: Just to double check, you're okay with moving all the method signatures to std::string including get and set attribute?

jjcherry56: I was suggesting changing the map so managing the deletes would be easier.
I think you can get away with changing the get/set methods and rely on automatic coercion when const char*'s are passed in. Sometime I just try it can see what feels cleaner.
So ya, I am ok with it if it doesn't end getting messy.

QuantamHD:
Thanks for the feedback.

Addressed all comments. After moving to std::string I needed to wrap it in std::optional in order to preserve an unset value check.

I the new API overall, but if you don't we can go back to const char * or return an empty std::string when the attribute is not found.

jjcherry56:
I am having a hard time seeing a reason to distinguish between an empty string and a missing value. I don't think the std::optional is very useful here. Please use a "using" to get rid of all the std::'s. I think string is already covered by StringUtil.hh

QuantamHD:
I think the question is, is this a valid attribute `(* src="" *)`, and if it is can an API user differentiate between `(* src="" *)` and `(**)`. The optional would be empty in the latter, but not the former.

I think I tend to agree with you though. It seems like in most other place we don't differentiate between empty string and not present. I'll go back to just `string`

QuantamHD:
All feedback has been addressed.

jjcherry56 (approved):
ship it

QuantamHD:
Cool, Merge whenever you're ready. I don't have any permission on this repo

akashlevy:
Hey Ethan,

I am finding that this PR results in a parsing error when >2 attributes are present on a line:
```
netlist.v line 1520, syntax error, unexpected ',', expecting ATTRIBUTE_CLOSED
```

To reproduce, try a multi-attribute statement like: `(* src = "design.v:1.1-2.1", test1 = "test1", test2 = "test2" *)`

PR #18, opened 2024-03-27
Simple and flat `ArrayTable` and `ObjectTable`

kbieganski:
This patch simplifies the implementation of `ArrayTable` and `ObjectTable`.
Instead of being arrays of pointers to pointers, they are now flattened into arrays of pointers.
Also, `ArrayTable` now uses `std::vector`, leading to less manual memory management and overall less manual bookkeeping.
I expect this to bring a slight speedup, but I haven't measured it properly yet.

### DIFF
--- a/include/sta/ArrayTable.hh
+++ b/include/sta/ArrayTable.hh
@@ -58,7 +58,7 @@ public:
 
 private:
   ArrayBlock<TYPE> *makeBlock(uint32_t size);
-  void pushBlock(ArrayBlock<TYPE> *block);
+  void pushBlock(size_t size);
   void deleteBlocks();
 
   size_t size_;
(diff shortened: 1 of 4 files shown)


kbieganski:
I dropped the change to `ObjectTable`.

rovinski:
Do you have any benchmark measurements to show improvement? I think it is unlikely to be approved unless there are.

kbieganski:
The goal of this PR was to simplify the code, but I observed about 1% speedup on BlackParrot and 2% on Ibex.

include/sta/ArrayTable.hh
```diff
@@ -58,19 +58,16 @@ public:
...
-  void pushBlock(ArrayBlock<TYPE> *block);
+  void pushBlock(size_t size);
   void deleteBlocks();
 
   size_t size_;
   // Block index of free block (blocks_[size - 1]).
   BlockIdx free_block_idx_;
   // Index of next free object in free_block_idx_.
   ObjectIdx free_idx_;
-  // Don't use std::vector so growing blocks_ can be thread safe.
```
jjcherry56, line 69: I guess you missed the comment that you deleted that explains why std::vector cannot be used here. If you look at the commit history you will see that it used to use std::vector. When the the vector grows it fails with multiple threads.

kbieganski: I didn't miss it. `vector::swap` is basically no different from the previous approach. See https://godbolt.org/z/8sqqTPMMf

However, I suppose implementations other than libstdc++ might exhibit different behavior on lower optimization levels. So I pushed a manual swap implementation just to be sure.

kbieganski:
@jjcherry56 Do you have any further feedback?

parallaxsw (commented):
I don't really see a compelling reason to make these changes. It is only marginally "simpler".  The last change to replace swapping the vectors is a giant step backwards.

Issue #19, opened 2024-03-27
BUG: attribute parsing with >2 attributes


akashlevy:
I am finding that PR #11 results in a parsing error when >2 attributes are present on a line:

`netlist.v line 1520, syntax error, unexpected ',', expecting ATTRIBUTE_CLOSED`

To reproduce, try a multi-attribute statement like: `(* src = "design.v:1.1-2.1", test1 = "test1", test2 = "test2" *)`

@QuantamHD 

I'm happy to try to help write a fix.

QuantamHD:
Yeah give it a shot. I think it's probably an easy fix. The internal data structures support it, so probably just need to update the grammar file.

If you're having a hard time fixing it let me know. Should be somewhere in this area https://github.com/parallaxsw/OpenSTA/blob/master/verilog/VerilogParse.yy#L485-L522

akashlevy:
Thanks Ethan. I'll give it a shot some time this week or next.

QuantamHD:
I'm away from my work computer for the rest of the day, but I think this is actually just a one character change if you want to make a quick PR.

I think line https://github.com/parallaxsw/OpenSTA/blob/master/verilog/VerilogParse.yy#L502 should read `attr_specs ',' attr_spec` not `attr_spec ',' attr_spec`

Also if you could update the test https://github.com/parallaxsw/OpenSTA/blob/master/test/verilog_attribute.v to include three attributes that would be much appreciated.

akashlevy:
Good catch. That's got to be the reason why 2 attributes works fine, but 3 doesn't. I'm working on the PR right now.

akashlevy:
Resolved with #20

PR #20, opened 2024-03-28
Fix parsing error with >2 attributes

akashlevy:
Addressing issue #19, introduced with #11. Credits to @QuantamHD for identifying the fix.

0-2 attributes could be parsed correctly, but 3+ produced an error. This was due to a typo in the recursive attribute parsing of the verilog yacc file.

Fixed the error, added a test case with 3 attributes. Problem appears to be resolved now, contributing back upstream.


QuantamHD:
Thanks @akashlevy! 

@jjcherry56 could you take a look when you have a free moment?

jjcherry56 (commented):
looks good

PR #22, opened 2024-04-10
Parallel `findClkSkew`

kbieganski:
Runs `findClkSkew` in parallel.

`findClkSkew` basically retrieves the absolute maximum skew per clock.
We can do it in separate threads and then reduce over the partial results from each thread.

## Benchmarks

Ibex
BlackParrot


kbieganski:
@jjcherry56 Can you take a look?

PR #50, opened 2024-07-18
Support JSON format and redirect to file for `report_units`

akashlevy:
- Add a `report_units -json` flag to support unit reporting in JSON format
- Add support for redirect to file


akashlevy:
@jjcherry56 let me know if you have interest in upstreaming this. If so, I can add it to the doc as well

jjcherry56:
I can't imagine why it would be useful to print json to the console instead of a file, so it only makes sense when used with redirection. I don't know the context of why json for units is useful, but suggest defining a tcl proc sta::write_unit_json filename instead.

akashlevy:
Yes, this is only useful when other tools need to ingest the units for post-processing. I can move this to a separate tcl proc.

akashlevy:
Just gonna make a new PR

PR #60, opened 2024-07-29
Refactor `filter_*` and add support for `!~`

akashlevy:
- Refactor `Sdc.i` to use templates for `filter_*` functions
- Add support for "not exact match" in filter expressions, represented by `!~` operator
- Update docs accordingly


sdc/Sdc.i
```diff
@@ -1270,30 +1270,41 @@ all_outputs_cmd()
...
-filter_ports(const char *property,
-	     const char *op,
-	     const char *pattern,
-	     PortSeq *ports)
-{
-  PortSeq filtered_ports;
-  if (ports) {
+template <typename T> Vector<T*>
+filter_objects(const char *property,
+         const char *op,
```
jjcherry56, line 1275: indentation

sdc/Sdc.i
```diff
@@ -1270,30 +1270,41 @@ all_outputs_cmd()
...
     Sta *sta = Sta::sta();
     bool exact_match = stringEq(op, "==");
     bool pattern_match = stringEq(op, "=~");
     bool not_match = stringEq(op, "!=");
-    for (const Port *port : *ports) {
-      PropertyValue value(getProperty(port, property, sta));
+    bool not_pattern_match = stringEq(op, "!~");
+    for (T *object : *objects) {
+      PropertyValue value(getProperty(object, property, sta));
       const char *prop = value.stringValue();
```
jjcherry56, line 1288: this has to be asString; stringValue only applies to properties that have string value (stringValue() should probably error to check)

sdc/Sdc.i
akashlevy: Sorry my VS code settings were wrong...

sdc/Sdc.i
akashlevy: Fixed

PR #64, opened 2024-08-01
Create partial set of one-to-one timing arcs for Liberty cell buses when input/output widths are mismatched

akashlevy:
In Liberty, a "one-to-one" timing arc set is created if an output bus function is defined using an unindexed input bus. For example, suppose we have a 4b-wide inverter with `A[3:0]` and `Y[3:0]`. The function for `Y` would be `!A` and this would result in a one-to-one timing arc set: `A[3]->Y[3]`, `A[2]->Y[2]`, `A[1]->Y[1]`, `A[0]->Y[0]`.

What happens in OpenSTA if `A` and `Y` have different widths? Before this PR, a warning would be shown and no timing arcs created. This can be seen by running the included `one2one` test case.

Sometimes having different widths is intended behavior and we want to create timing arcs across the pins that do match. For example if `A` is width 8 and `Y` is width 4 , then we could warn the user, but still create timing arcs between `A[3]->Y[3]`, `A[2]->Y[2]`, `A[1]->Y[1]`, `A[0]->Y[0]`. This new behavior is implemented in this PR (along with test cases).

### DIFF
--- a/include/sta/Liberty.hh
+++ b/include/sta/Liberty.hh
@@ -556,7 +556,10 @@ protected:
                                const LibertyPort *en,
                                const RiseFall *en_rf,
                                const LibertyPort *q,
+                               const TimingArcSet *en_to_q,
                                Report *report);
+  bool condMatch(const TimingArcSet *arc_set1,
+                 const TimingArcSet *arc_set2);
   void findDefaultCondArcs();
   void translatePresetClrCheckRoles();
   void inferLatchRoles(Report *report,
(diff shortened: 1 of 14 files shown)


jjcherry56 (commented):
Looks good.
Please rename the files to liberty_arcs_one2one.tcl/lib/v. I never include the word "test" in test filenames because it is implied and I use liberty as a prefix for liberty regressions (not publicly visible).

akashlevy:
Done

PR #67, opened 2024-08-05
CentOS/RHEL 7 Dockerfile

akashlevy:
This PR improves the CentOS 7 compatibility in the `Dockerfile` and `CMakeLists.txt`:
- Add `stdc++fs` as a target link library as required for older GCC versions needing explicit C++17 `filesystem` support
- Switch to using `centos-release-scl` and `epel-release` repos for all packages except CUDD
- Adding support for CentOS 7 Vault repos (since CentOS 7 is officially EOL now) using recommended `sed` commands
- Fix CUDD download link

Also:
- Fix minor typo in `CMakeLists.txt`


jjcherry56 (commented):
Wouldn't it make more sense to update the docker base to use a newer version of centos instead of patching OpenSTA to support a version on EOL?

akashlevy:
Even though CentOS 7 is EOL, RHEL7 isn't EOL until 2028. This seemed to me the best way to keep RHEL7 compatibility, which a lot of people are still using.

CMakeLists.txt
```diff
@@ -509,6 +509,7 @@ target_sources(OpenSTA
 )
 
 target_link_libraries(OpenSTA
+  stdc++fs
```
jjcherry56, line 512: this has to be conditionalized because it cause errors:
ld: library 'stdc++fs' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)

akashlevy: I'll fix

jjcherry56:
I don't understand what all the sed commands are about. 
How are they "recommended"?
I will trust you on this because I try and steer clear of docker.

akashlevy:
The `sed` stuff is just to switch to using the CentOS vault repo instead of the main CentOS repo, which no longer hosts packages since EOL

akashlevy:
"Recommended" should probably be used in quotes, because it is what the top 3 Google search results yielded me when I looked up the errors I was getting accessing main repo

jjcherry56:
It seems like this should update the docker file name to Dockerfile.CentOS7.

PR #69, opened 2024-08-07
search: allow path groups updating to be triggered externally

AcKoucher:
Required for the approach in [#5431](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5431) so that `path_groups_` can  be populated without having to call `Search::findPathEnds`


jjcherry56 (commented):
see 32830ef5 Search::makePathGroups set path_groups_
for a much simpler way to do this

maliberty:
Search::makePathGroups is protected so we can't call it.  Please make it public.

PR #76, opened 2024-08-13
Sdc.tcl: invoke original_source if we can

oharboe:
This preserves behaviors such as 'info script' of the original tcl implementation.


oharboe:
@tspyrou @maliberty FYI

jjcherry56:
This fails a large number of regressions in opensta and openroad (try rsz/test/regression for example).
The issue is bindings of variables set in a sourced script don't stick.

oharboe:
@jjcherry56 Thanks for the feedback. I will try again if I find an approach tht works with rsz/test/regression

jjcherry56:
I tried a few things on my end as well. I couldn't find anything.

PR #83, opened 2024-09-03
add cell footprint support & add it to equiv check

AcKoucher:
This is to address some [Resizing Problems](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5577) in OR. With this, RSZ won't swap cells that are functionally equivalent but non equivalent footprint wise.

### DIFF
--- a/liberty/EquivCells.cc
+++ b/liberty/EquivCells.cc
@@ -530,7 +530,7 @@ bool
 equivCellFootprints(const LibertyCell *cell1,
                     const LibertyCell *cell2)
 {
-  return strcmp(cell1->footprint(), cell2->footprint()) == 0;
+  return stringEqIf(cell1->footprint(), cell2->footprint());
 }
 
 } // namespace
--- a/liberty/EquivCells.cc
+++ b/liberty/EquivCells.cc
@@ -325,7 +325,8 @@ equivCells(const LibertyCell *cell1,
     && equivCellPgPorts(cell1, cell2)
     && equivCellSequentials(cell1, cell2)
     && equivCellStatetables(cell1, cell2)
-    && equivCellTimingArcSets(cell1, cell2);
+    && equivCellTimingArcSets(cell1, cell2)
+    && equivCellFootprints(cell1, cell2);
 }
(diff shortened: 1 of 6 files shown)


jjcherry56:
If you have an int -> string map there is no point in having the vector.
A better solution would be a set of strings and skip all the extraneous indexing code.
But I really don't see any point in trying to save a few strings the footprints.
The bulk of the memory in a liberty file is the tables, not something like a footprint
And if you do go to the trouble of using indices as a way to make unique strings, there is no point
in using strcmp to compare them when an integer compare would suffice.

OpenSTA does not use OpenROAD formatting for code and unfortunately clang-format does not work well enough to do it for you.
See OpenSTA/doc/CodingGuidelines or look at the other code in the file to see what the standard is.

AcKoucher:
@jjcherry56 Thanks for the feedback. I simplified it by just keeping the string within the cell and adapted the curly brackets to OpenSTA formatting. Please, let me know if there's more to address.

liberty/EquivCells.cc
```diff
@@ -525,4 +526,11 @@ equivCellTimingArcSets(const LibertyCell *cell1,
   }
 }
 
+bool
+equivCellFootprints(const LibertyCell *cell1,
+                    const LibertyCell *cell2)
+{
+  return strcmp(cell1->footprint(), cell2->footprint()) == 0;
```
jjcherry56, line 533: stringEq

liberty/Liberty.cc
```diff
@@ -1989,6 +1992,21 @@ LibertyCell::ensureVoltageWaveforms(const DcalcAnalysisPtSeq &dcalc_aps)
...
 
+void
+LibertyCell::setFootprint(const char *footprint)
+{
+  footprint_ = stringCopy(footprint);
+}
+
+const char*
+LibertyCell::footprint() const
+{
```
jjcherry56, line 2003: This should just return nullptr if there is no footprint instead of ""

AcKoucher:
Review addressed.

PR #85, opened 2024-09-07
build and run tests on each `Dockerfile.*`

gnawhleinad:
Irrespective of https://github.com/parallaxsw/OpenSTA/pull/84, this pull request adds GitHub Action tests to build and run from the `Dockerfile`s in this repository.

It looks like it's both systems are failing on `liberty_arcs_one2one`...?

PR #86, opened 2024-09-10
Smallfix to get `liberty_arcs_one2one` tests consistently passing in regression

akashlevy:
#85 highlighted that `liberty_arcs_one2one` fails under different machine configurations (via Docker). This PR fixes the issue by using the `summary` mode of `report_checks` to avoid having rising/falling edges be reported in a non-deterministic way.


jjcherry56:
I think in this case you may want to use report_edges -from to query the graph directly instead of using reporting.

akashlevy:
True, I'm still relying on equi-delay paths being reported in a certain order. Let me go fix that...

akashlevy:
Ok, I switched to using `report_edges`, but I'm seeing two potential issues...

1. I'm confused why there seem to be four timing arcs per edge instead of two? An inverter is negative unate, so I don't know why we are seeing something that looks non-unate...
2. Is the order in which the different arcs are reported a deterministic thing for `report_edges`?

akashlevy:
Got it. I've added the timing sense to the Liberty files as suggested, which resolved the issue with non-unate arcs.

I have split and renamed the test cases. Added comments to line 1 of all tests describing what they do. Cross-platform runs are going through on my Mac as well as the two Linux Dockers.

Let me know if anything else is needed. Might be helpful to have a set of test guidelines in `docs/CodingGuidelines.txt`, so contributors can follow that.

PR #89, opened 2024-09-12
Remove the footprint check from equivCells

maliberty:
This check will become part of OpenROAD and not sta to allow for more flexibility and augmentation with physical data.


maliberty:
@gadfort @QuantamHD FYI

maliberty:
This is a partial revert of #83

PR #90, opened 2024-09-12
Parse and store user_function_class from Liberty.

maliberty:
Added user_function_class and cell_footprint to LibertyWriter for testing.


maliberty:
@jjcherry56 any feedback?

PR #95, opened 2024-09-17
Bugfix: SDC square bracket escaping on single-bit input ports

akashlevy:
I've included a minimal test case where OpenSTA fails to annotate input delays on a valid netlist (try `verilog_square_bracket` test). It's a bit of a corner case, but I was ultimately able to isolate and resolve the problem.

It's related to naming of ports in SDC when square brackets exist in the middle of the port names. These square brackets need to be escaped. There is some code that already does this for bus ports, but that code doesn't handle single-bit ports.

I've essentially replicated all the places I found the escape logic to also handle single-bit port scenarios, and with that code added, the test case passes.

### DIFF
--- a/network/SdcNetwork.cc
+++ b/network/SdcNetwork.cc
@@ -664,6 +664,12 @@ SdcNetwork::findPort(const Cell *cell,
 	port = network_->findPort(cell, escaped2.c_str());
       }
     }
+    else {
+      // Try escaping pin name
+      string escaped2;
+      string escaped_pin_name = escapeBrackets(name, this);
+      port = network_->findPort(cell, escaped_pin_name.c_str());
+    }
   }
   return port;
 }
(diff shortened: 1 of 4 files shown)


jjcherry56:
You identified 3 different bugs in port lookup but the test only really checks one of them, albeit rather indirectly. These lines in the regression test the issues more directly:
report_object_name [get_ports in[0].test]
report_object_name [[sta::top_instance] find_pin in[0].test]
report_object_name [sta::find_pin in[0].test]
report_object_name [[[sta::top_instance] cell] find_port in[0].test]

I removed some dead code and made the variable names more consistent with the existing code and commited it as
7de90722 get_ports for verilog \foo[0].bar
I have an existing private regression test for verilog escaping that I updated to include this issue.

akashlevy:
Thanks Cherry

PR #99, opened 2024-09-22
Allow SDC `get_*` with 0 args and on object references

akashlevy:
This PR adds two features into SDC processing:
- `get_*` with no arguments is interpreted as `get_* *`
- `get_*` called on SWIG object references `*_p_<objtype>` will keep those references intact rather than failing

Let me know if anything should be added to the docs for these changes.


jjcherry56:
You need to type check the objects:
get_pins [get_cells]
_e001650300600000_p_Instance _8001650300600000_p_Instance _6000650300600000_p_Instance _2001650300600000_p_Instance _c000650300600000_p_Instance

I would prefer it if you did not use continue.
Please break the tests into 2 tests; one for no arg and the other for passing objects.

akashlevy:
Sure, will do these tonight

akashlevy:
Ok, made the changes:
- I added error messages for when the type is wrong and updated `messages.txt`
- Switched from using `continue` to `else { ... }`
- Split into test cases for `get_noargs` and `get_objrefs`

akashlevy:
Let me know if anything else is needed :)

PR #101, opened 2024-09-26
Add `src_attr` and `netlist_line` fields to `report_checks`

akashlevy:
See title. `src_attr` is the `src` attribute of an instance in the netlist (useful for tracking an object back to source in pre-synthesis RTL). `netlist_line` is the line number for tracking an instance back to its source in the Verilog netlist.

Breakdown of work:
- [x] Update `Sta`, `Network`, `ConcreteNetwork`, and `SdcNetwork` to include `line` (line number of instantiation)
- [x] Update `ReportPath` to add the new fields to reporters
- [x] Update `Search` to add TCL interface to use the new fields
- [x] Add test to check that the new fields are reported correctly
- [x] Update docs to include new fields

Note that I updated the `Sta` API to support getting/setting line numbers on `Instance`/`ConcreteInstance`. I made specifying the line number optional by providing a default value, but anything implementing `Network` still needs to implement the `line()` function right now...

Let me know if you have any suggestions on the right strategy for annotating line numbers on instances.


jjcherry56:
I understand why the src_attr is useful, but why is the netlist line interesting?

akashlevy:
To be able to track where an instance is instantiated in the netlist? Feels kind of useful to me, e.g. when you see an unexpected cell type on a critical path. But maybe not. Happy to upstream just `src_attr` if you prefer.

QuantamHD:
How does this interact with the (* src = "file.v:245" *) attributes emitted by Yosys? Should we somehow leverage the existing infrastructure for that rather than add new fields?

akashlevy:
@QuantamHD It is leveraging the existing infrastructure for `src` attribute parsing that you added. So not quite sure what you mean

QuantamHD:
Ah, my bad @akashlevy. I didn't read it carefully enough

akashlevy:
@jjcherry56 wanted to give this a bump. Any feedback?

akashlevy:
Going to close this and re-open with just `src_attr` soon as discussed.

PR #110, opened 2024-10-20
Hierarchical Verilog writer repairs

andyfox-rushc:
Hi,
We observed some strange behaviour with the iterators on pending_children_ during writeModule (we saw a seg fault). Matt suggested replacing the vector with a deque which seems to do the trick. I am not quite sure why the original code failed (looked ok to me) but the problem seemed to occur when at the last element of the pending_children_ and then the recursive call adding another element.
The fix includes the replacement of the pending_children_  Vector -> deque in writeModule in VerilogWriter.cc
Also, I added a sort for deque in Vector.hh (am guessing there might usefully be a new header file wrapper for deque).

### DIFF
--- a/include/sta/Vector.hh
+++ b/include/sta/Vector.hh
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <vector>
+#include <deque>
 #include <algorithm>
 
 namespace sta {
@@ -126,7 +127,7 @@ sort(Vector<OBJ> &seq, SortCmp cmp)
   std::stable_sort(seq.begin(), seq.end(), cmp);
 }
 
-template <class OBJ, class SortCmp>
+  template <class OBJ, class SortCmp>
 void
 sort(Vector<OBJ> *seq, SortCmp cmp)
 {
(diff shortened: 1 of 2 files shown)


parallaxsw:
The failure was caused by the range iteration on the pending_children_ with recursive calls that added to the vector invalidating the iterator. There is another issue that sorting and collect the modules as they are being written doesn't make much sense. I pushed 

1dbcb329 write_verilog separate cell search from write

that separates the search from the writing that should fix your issue.

PS, it also doesn't make any sense to add a sort template for deque to Vector.hh.

maliberty:
@andyfox-rushc is the above commit sufficient to resolve your issue and close this PR?  If so I can update sta in OR.

andyfox-rushc:
While James is very kindly looking at his code (I note he caught the weird sorting issue -- where at every level of recursion all the children were being sorted), please can he also clean up the findUnconnectedNetCount ? This is also another little issue. Specifically it is being called by writeWireDcls (called by the recursive call writeModule)  for every instance and it traverses the whole hierarchy.  

VerilogWriter::findUnconnectedNetCount()
{
  return findNCcount(network_->topInstance()); <-- no need to do the whole hierarchy !!
}

Probably best to to keep this PR open until Cho hash checked out a few more hierarchical cases. For example the uses of bus ports at intermediate levels may have some weirdness.

My understanding is that the OpenROAD verilog writer was primarily used on flat designs after the flow has been run so it is quite reasonable to expect there to be a few more little things in the hierarchical writing.

Thank you James for looking into these things.

parallaxsw:
Independent of what further testing reveals this PR is no longer relevant so I am going to close it.
The issue with findUnconnectedNetCount is actually that the unconnected pins are always counted for the the top level module (there is no recursive call). I addressed that issue with
a2d445b0  write_verilog unconnected wire dcls for non-top level modules

Issue #114, opened 2024-10-30
BUG: CCSN Liberty parsing with `:` in `input_ccb` group


akashlevy:
Had a customer report an OpenSTA parsing issue with a commercial Liberty file. The problematic line looks like this (some PDK identifiers have been obfuscated here):
```
input_ccb(ECOMX2_D1_N_XXXXXXX_XXXX__XXX_1_4_XXXXX_XXXXX:a) {
```
Looks like a CCSN group. The customer is working on providing us an obfuscated test case, which I will attach.

I assume the fix is pretty straightforward: allow colons in groups? Let me know if you'd like me to try to make a PR for this @jjcherry56. Also let me know if there are any things I need to be aware of with allowing colons (maybe some other groups treat colons in a special way? Not sure)

akashlevy:
```
library (broken_ccsn_colon) {
  delay_model : "table_lookup";
  simulation : false;
  capacitive_load_unit (1,pF);
  leakage_power_unit : "1pW";
  current_unit : "1A";
  pulling_resistance_unit : "1kohm";
  time_unit : "1ns";
  voltage_unit : "1v";
  library_features : "report_delay_calculation";
  input_threshold_pct_rise : 50;
  input_threshold_pct_fall : 50;
  output_threshold_pct_rise : 50;
  output_threshold_pct_fall : 50;
  slew_lower_threshold_pct_rise : 30;
  slew_lower_threshold_pct_fall : 30;
  slew_upper_threshold_pct_rise : 70;
  slew_upper_threshold_pct_fall : 70;
  slew_derate_from_library : 1.0;
  nom_process : 1.0;
  nom_temperature : 85.0;
  nom_voltage : 0.75;

  cell(obfuscated) { 
    sensitization_master : sensitization_3pins ; 
    area : 0.1 ; 
    cell_footprint : obfuscated_D2_N ; 
    dont_touch : true ; 
    dont_use : true ; 

    pin(A) { 
      capacitance : 0.0001 ; 
      direction : input ; 
      driver_waveform_rise : "driver_waveform_default_rise" ; 
      driver_waveform_fall : "driver_waveform_default_fall" ; 
      fall_capacitance : 0.0001 ; 
      input_voltage : default ; 
      max_transition : 0.1 ; 
      related_ground_pin : VSS ; 
      related_power_pin : VDD ; 
      rise_capacitance : 0.0001 ; 

[...]

PR #115, opened 2024-10-31
BUGFIX: Adjust Liberty parser to accept groups with colons

akashlevy:
See issue #114 for more details. Quick (conservative) fix by adding more permissiveness on the group name with `colon_expr` in `liberty/LibertyParse.yy`. Also added the testcase in #114 to check that parsing of the Liberty file doesn't fail.

### DIFF
--- a/liberty/LibertyParse.yy
+++ b/liberty/LibertyParse.yy
@@ -50,7 +50,7 @@ int LibertyLex_lex();
 %token <string> STRING KEYWORD
 
 %type <stmt> statement complex_attr simple_attr variable group file
-%type <attr_values> attr_values
+%type <attr_values> attr_values colon_expr
 %type <attr_value> attr_value
 %type <string> string expr expr_term expr_term1 volt_expr
 %type <line> line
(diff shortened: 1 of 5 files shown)


QuantamHD:
Does the liberty spec say what's allowed here out of curiosity?

QuantamHD:
I looked through a couple of manuals and it seems like it's left completely unspecified what's allowed in there other than "string". 😞 

https://media.c3d2.de/mgoblin_media/media_entries/659/Liberty_User_Guides_and_Reference_Manual_Suite_Version_2017.06.pdf

I'm guessing pretty much anything is allowed in there other than ( and ).

akashlevy:
(I wish the semiconductor industry would just switch from Liberty to JSON like sane people)

QuantamHD:
(\(#metoo\))

akashlevy:
fixed by a 1 character change in
https://github.com/parallaxsw/OpenSTA/commit/d889f03757d0394cacf1bbdf0b4fb203d4b066f3

Closing

PR #117, opened 2024-11-04
Silence unused warnings on StaTclTypes.i helpers

povik:
As a follow-up to #62, we need to silence unused warnings for when `StaTclTypes.i` is included in an external SWIG library (like openroad's resizer) and only a subset of the helpers are used.

`[[maybe_unused]]` is a C++17 extension but the OpenSTA build is configured for C++17 so this should be fine.

### DIFF
--- a/tcl/StaTclTypes.i
+++ b/tcl/StaTclTypes.i
@@ -193,7 +193,7 @@ tclListNetworkSet1(Tcl_Obj *const source,
   return set;
 }
 
-static StringSet *
+[[maybe_unused]] static StringSet *
 tclListSetConstChar(Tcl_Obj *const source,
 		    Tcl_Interp *interp)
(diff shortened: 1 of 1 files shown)


akashlevy:
+1 from us. We have a similar thing in https://github.com/Silimate/silisizer to avoid these warnings.

parallaxsw:
Since the functions are called from multiple files in that situation I'd rather see them migrated to non-static functions defined in a file named StaTclTypeHelpers.cc/.hh

akashlevy:
@jjcherry56 this can be closed now that #182 is merged

@povik if you take the latest version, just include `TclTypeHelpers.hh` to use the utils--the warnings should be gone

povik:
Cool!

PR #118, opened 2024-11-04
Add warning for poor activities convergence

povik:
Please consider having a warning like this.

I've found on some designs in OpenROAD-flow-scripts the propagated activities don't converge mostly due to bad features in the input netlist. Having a warning like this would help uncover these cases.

I've picked an arbitrary value of 0.20 for the max change threshold, and as for the mesage number I picked one above the existing Power.cc message. Not sure if `messages.txt` is maintained by some automated tooling but I tried inserting the message manually.

### DIFF
--- a/power/Power.cc
+++ b/power/Power.cc
@@ -629,22 +629,28 @@ Power::ensureActivities()
       bfs.visit(levelize_->maxLevel(), &visitor);
       // Propagate activiities through registers.
       InstanceSet regs = std::move(visitor.visitedRegs());
-      int pass = 1;
-      while (!regs.empty() && pass < max_activity_passes_) {
-        visitor.init();
-        InstanceSet::Iterator reg_iter(regs);
-	while (reg_iter.hasNext()) {
-	  const Instance *reg = reg_iter.next();
-	  // Propagate activiities across register D->Q.
-	  seedRegOutputActivities(reg, bfs);
-	}
-	// Propagate register output activities through
-	// combinational logic.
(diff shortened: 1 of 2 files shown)


akashlevy:
@povik (For future reference, `etc/FindMessages.tcl` regenerates `docs/messages.txt`)

power/Power.cc
```diff
@@ -629,22 +629,28 @@ Power::ensureActivities()
...
+          }
+          // Propagate register output activities through
+          // combinational logic.
+          bfs.visit(levelize_->maxLevel(), &visitor);
+          regs = std::move(visitor.visitedRegs());
+          debugPrint(debug_, "power_activity", 1, "Pass %d change %.2f",
+                     pass, visitor.maxChange());
+          pass++;
+        }
+        if (visitor.maxChange() > 0.20) {
```
parallaxsw, line 650: This warns for nearly every power regression I have even though the activities indeed do converge. Consider a circuit with no feedback. When the activities propagate on the first pass they change by whatever the input activity is (with may be > .2). Since there is no feedback they are not re-queued and regs is empty. The activities are converged but it still warns about non-convergence. I think the true criteria for non-convergence is  "pass == max_activity_passes_". All of the opensta regressions I have converge, so maybe the place to start is to make a test case and submit it as an issue.

povik: Yes, the suggested condition is wrong. I'll open a new PR if I revisit this

PR #122, opened 2024-11-07
Add <cstdlib> header for later cpp compatibility.

mikesinouye:
std::abs has been moved to <cstdlib> in later versions of the standard.

https://en.cppreference.com/w/cpp/numeric/math/abs

### DIFF
--- a/dcalc/FindRoot.cc
+++ b/dcalc/FindRoot.cc
@@ -17,6 +17,7 @@
 #include "FindRoot.hh"
 
 #include <algorithm> // abs
+#include <cstdlib> // abs for C++11 
 
 namespace sta {
 
--- a/dcalc/FindRoot.cc
+++ b/dcalc/FindRoot.cc
@@ -16,8 +16,7 @@
 
 #include "FindRoot.hh"
 
-#include <algorithm> // abs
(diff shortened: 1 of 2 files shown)


util/StringUtil.cc
```diff
@@ -19,6 +19,7 @@
 #include <limits>
 #include <cctype>
 #include <cstdio>
+#include <cstdlib>
```
parallaxsw, line 22: Please explain what function this meant to include.

dcalc/FindRoot.cc
```diff
@@ -17,6 +17,7 @@
 #include "FindRoot.hh"
 
 #include <algorithm> // abs
+#include <cstdlib> // abs for C++11
```
parallaxsw, line 20: It looks to me like including just cmath makes more sense than including both cstdlib and algorithm.

util/StringUtil.cc
mikesinouye: Added.

dcalc/FindRoot.cc
mikesinouye: Good point, fixed.

PR #123, opened 2024-11-08
Add the signal_type info to the test_cell pins.

maliberty:
Now that we have different pins on the cell and the test_cell we should put the information on both.

### DIFF
--- a/liberty/LibertyReader.cc
+++ b/liberty/LibertyReader.cc
@@ -97,6 +97,7 @@ LibertyReader::init(const char *filename,
   op_cond_ = nullptr;
   ports_ = nullptr;
   port_ = nullptr;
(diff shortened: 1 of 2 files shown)


maliberty:
Any comments?

parallaxsw:
It should use the existing ports_ var instead of introducing another one.
See 0d138393 liberty test_cell port signal_type

maliberty:
Did you mean the `port_` var?  It points at a different port:
```
      port_ = findPort(save_cell_, pin_name);
      test_port_ = findPort(test_cell_, pin_name);
```
Having the "same" port on the cell & test_cell introduces the need for two variables.

maliberty:
I see you've changed the implementation since this PR.  I'll review and update.

PR #129, opened 2024-11-14
Make `is_memory` property more sensitive for cells and libcells by also matching on `memory` groups

### LINKED ISSUE #124: `is_memory` property could be more sensitive
For a Liberty cell, `is_memory` is only true if `is_memory : true` attribute is on the cell. But I've found many cases in the wild where that attribute is missing for RAMs/ROMs. Instead, these cells have a `memory () { ... }` group, which definitively marks them as being memories. Would it make sense to check for `memory () { ... }` as another possible condition for the `is_memory` property?

Let me know if you'd like me to implement and PR this.

akashlevy:
See title. Resolves #124. 

- [x] `doc/OpenSTA.odt`, `doc/OpenSTA.pdf`: Update docs **(TODO after #128 is upstreamed)**
- [x] `liberty/LibertyReader.cc`, `liberty/LibertyReaderPvt.hh`: Liberty group visitor for `memory ()`
- [x] `search/Property.cc`: Add `is_memory` property to instances and Liberty cells in addition to `is_memory_cell`
- [x] `test/gf180mcu_sram.lib.gz`: Commercial open-source example of RAM with `memory ()` group, but no `is_memory` attribute
- [x] `test/get_is_memory.tcl`, `test/get_is_memory.v`, `test/get_is_memory.ok`, `test/regression_vars.tcl`
  - [x] Add test case for filtering memory instances
  - [x] Add test case for filtering memory Liberty cells


test/get_has_memory.tcl
```diff
@@ -0,0 +1,11 @@
+# Tests whether the has_memory attribute works for cells and libcells
+read_liberty ../examples/gf180mcu_sram.lib.gz
```
parallaxsw, line 2: this lib is not in the examples dir

akashlevy: Sorry, I'm confused. https://github.com/parallaxsw/OpenSTA/pull/129/files#diff-71319827d46ce51b81049eef16dce8bb178cc42ebdfd540e0bc8673ad24379c1

Do you want me to move it to the test directory?

include/sta/Liberty.hh
```diff
@@ -591,6 +593,7 @@ protected:
   LibertyLibrary *liberty_library_;
   float area_;
   bool dont_use_;
+  bool has_memory_;
```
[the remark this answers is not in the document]
akashlevy: Sorrry, fixed

test/get_has_memory.tcl
parallaxsw: weird, it doesn't show up in the patch file. but I think it would be better in the test dir because it isn't used by an example

akashlevy: Sure I can move it rn

akashlevy:
@jjcherry56 ready for review having switched to `is_memory`. Let me know if any more changes are needed

search/Property.cc
```diff
@@ -957,7 +960,8 @@ getProperty(const Instance *inst,
     return PropertyValue(liberty_cell && liberty_cell->isInverter());
   else if (stringEqual(property, "is_macro"))
     return PropertyValue(liberty_cell && liberty_cell->isMacro());
-  else if (stringEqual(property, "is_memory_cell"))
+  else if (stringEqual(property, "is_memory")
+	   || stringEqual(property, "is_memory_cell"))
```
parallaxsw, line 964: I am not sure why the original code used "is_memory_cell" but I think we should just flush it unless you see a use for it.

akashlevy: Ok

PR #130, opened 2024-11-19
Make MinMax/SetupHold/EarlyLate const

maliberty:
Makes clear that these are immutable objects.


maliberty:
In OR we are working to remove all non-const statics that could cause problems with having separate instances running in different threads.  While MinMax isn't strictly a problem it shows up as it is non-const, while being const in effect.

This PR clarifies it by actually making the objects const.  If you are agreeable to this change I will make follow on PRs for other similar value-wrapper types.

tspyrou:
This would ensure to the caller that these are unmodifiable in threads. We have users who are driving us in this direction.

maliberty:
No response, giving up.

PR #131, opened 2024-11-22
Rename flags in `report_checks`

### LINKED ISSUE #126: Suggestion: rename `-group_count` and `-endpoint_count` in `report_checks`
The naming for `-group_count` and `-endpoint_count` in `report_checks` is a bit unintuitive.

To me, the flag names seem to suggest that `-group_count` would set *the total number of path groups that get reported*, and `-endpoint_count` would set *the total number of endpoints that get reported*. Instead, `-group_count` sets the number of paths per path group and `-endpoint_count` sets the number of paths per endpoint.

I would consider renaming these flags to avoid end user confusion. Perhaps something more explicit like `-group_count`->`-paths_per_group` and `-endpoint_count`->`-paths_per_endpoint`.

Of course, you would probably want to still accept `-group_count` and `-endpoint_count` for backwards-compatibility.

Just a small suggestion. Let me know your thoughts.

akashlevy:
Resolves #126

- [x] Find and replace `group_count` with `group_path_count` globally in repo
- [x] Find and replace `endpoint_count` with `endpoint_path_count` globally in repo
- [x] Ensure that replacements result in <= 80 char lines
- [x] `search/Search.tcl`: Add new flag names, deprecation warning on old flag names
- [x] `doc/messages.txt`: Add new deprecation warnings and update line numbers
- [x] `doc/OpenSTA.odt`, `doc/OpenSTA.pdf`: Update flags in docs


akashlevy:
Here's the test case I used to validate. Let me know if I should add it to the `test/` directory in a commit. I assume you already have some `report_checks` test cases which it might be redundant with, so I didn't include it.

```
# report_checks with -group_path_count and -endpoint_path_count
read_liberty ../examples/sky130hd_tt.lib.gz
read_verilog verilog_attribute.v
link_design counter
create_clock -name clk -period 10 clk
set_input_delay -clock clk 0 [all_inputs -no_clocks]

puts {report_checks -endpoint_count 2}
report_checks -endpoint_count 2
puts {report_checks -group_count 3}
report_checks -group_count 3
puts {report_checks -endpoint_count 4 -group_count 5}
report_checks -endpoint_count 4 -group_count 5
puts {report_checks -endpoint_path_count 6}
report_checks -endpoint_path_count 6
puts {report_checks -group_path_count 7}
report_checks -group_path_count 7
```

search/Search.tcl
```diff
@@ -158,21 +159,31 @@ proc find_timing_paths_cmd { cmd args_var } {
...
-  set endpoint_count 1
-  if [info exists keys(-endpoint_count)] {
-    set endpoint_count $keys(-endpoint_count)
-    if { $endpoint_count < 1 } {
-      sta_error 512 "-endpoint_count must be a positive integer."
-    }
+  set endpoint_path_count 1
+  if { [info exists keys(-endpoint_count)] } {
+    # deprecated 2024-11-22
+    sta_warn 502 "find_timing_paths -endpoint_count is deprecated. Use -endpoint_path_count instead."
```
parallaxsw, line 165: The warnings should be for the command name $cmd, not the internal function.

akashlevy: Sorry, will fix

PR #135, opened 2024-12-05
Add {`instance`, `cell`, `verilog_src`, `pin`, `net`, `hier_pins`, `capacitance`} fields to `report_checks -format json` for paths

akashlevy:
Added more fields to `report_checks -format json` paths:
- [x] `instance`: Instance path name (if instance present)
- [x] `cell`: Cell name (if instance present)
- [x] `verilog_src`: Instance's Verilog src attribute (if instance present)
- [x] `pin`: Pin name
- [x] `net`: Name of net connected to pin (if valid net present)
- [x] `hier_pins`: Hierarchical pins (if pin is at module interface)
- [x] `capacitance`: Capacitance of pin (if pin is driver)

This PR also ensures that backslashes are escaped by using `sdc_network_` instead of `network_`, otherwise invalid JSON can be produced. It also refactors path end reporting to be 100% on the C++ side.

List of work:
- [x] `search/Sta.cc`, `include/sta/Sta.hh`: Add `reportPathEnds` function that calls `ReportPath::reportPathEnds`
- [x] `search/ReportPath.cc`: Add new fields, fix `reportPathEnds`, moved `hierPinsAbove` declarations up
- [x] `search/ReportPath.hh`: Add `pathFormat` getter
- [x] `tcl/StaTclTypes.i`: Add input type for `PathEndSeq*`
- [x] `search/Search.i`: Add `report_path_ends` wrapper function
- [x] `search/Search.tcl`: Remove `report_path_ends` TCL function (this has been SWIGified in `Search.i`)
- [x] `test/verilog_attribute.v`: Modify test case to be more "interesting" using two cells with intermediary net
- [x] `test/test_report_checks_src_attr.ok`: Update for change to `test/verilog_attribute.v` above
- [x] `test/report_json1.*`: Test case demoing new JSON fields
- [x] `test/report_json2.*`: Test case checking that no-paths condition still produces valid JSON
- [x] `test/regression_vars.tcl`: Add new test cases to regression


akashlevy:
Sorry I'm finding a few issues. I'm making this a draft PR until I'm confident this works on more complex cases.

jjcherry56:
please rename the "inst" key to "instance" and "src" to "source"; I don't see the need to abbreviate them

jjcherry56:
cell and liberty_cell are two different things but I think cell should be preferred and work even if there is no liberty cell.

I see your point on "src" but "src_attribute" is unfortunate and redundant. It should be something like verilog_src at least.

akashlevy:
`verilog_src` and keeping `cell` sounds good to me. What would be the situation in which there is no Liberty cell? A blackbox? What should `cell` show then?

((I'm also still working on validating this on some larger designs, which might take me some time. I also want to include `load_cap` as a field, as this as often quite useful for debugging delay values with NLDM.))

jjcherry56:
An instance always has a cell (network->cell(instance) which always has a name network->name(cell).
Yes it is a black box.

akashlevy:
@jjcherry56 this is ready for review now

search/ReportPath.cc
```diff
@@ -1086,9 +1095,9 @@ ReportPath::reportJson(const PathEnd *end,
   const Pin *startpoint = expanded.startPath()->vertex(this)->pin();
   const Pin *endpoint = expanded.endPath()->vertex(this)->pin();
   stringAppend(result, "  \"startpoint\": \"%s\",\n",
-               network_->pathName(startpoint));
+               escapeBackslashes(network_->pathName(startpoint)).c_str());
```
jjcherry56, line 1098: It doesn't work to use .c_str() with a temporary string returned by escapeBackslashes.
You would have to create a local var and then use .c_str() on it.
But I think the better solution is to write the names out in the sdc namespace by using sdc_network_ instead of network_.

test/regression_vars.tcl
```diff
@@ -132,6 +132,8 @@ record_sta_tests {
   get_objrefs
   get_lib_pins_of_objects
   report_checks_src_attr
+  report_checks_json
```
jjcherry56, line 135: how about report_json1 and report_json2 instead with a comment in the first line describing the test? that is the standard I use to manage 1100 regressions

akashlevy: Certainly can do.

search/ReportPath.cc
akashlevy: I'm not sure I understand the problem with using `c_str` against the `std::string` that is returned by `escapeBackslashes`. It appears to work great in the test case I have.

I can try using SDC namespace though; I assume that SDC namespace escapes backslashes automatically

akashlevy:
@jjcherry56 Made requested changes.

search/ReportPath.cc
jjcherry56: The sdc namespace does not escape hierarchy dividers that in an escaped verilog name.

akashlevy: Is backslash escaping needed?

akashlevy:
I am still confused on whether backslash escaping is needed. Please let me know

jjcherry56:
backslash escaping is not needed.

as for using c_str() returned strings, it works sometimes since the string is on the stack it is poped before the .c_str() is called so it fails. I wish it worked; it makes usiing strings with printf much more
painful.

akashlevy:
Ok removed backslash escaping again. Sorry about the confusion

search/Search.tcl
```diff
@@ -424,7 +424,11 @@ proc_redirect report_checks {
   parse_report_path_options "report_checks" args "full" 0
   set path_ends [find_timing_paths_cmd "report_checks" args]
   if { $path_ends == {} } {
-    report_line "No paths found."
+    if { $format == "json" } {
+      report_line "{\"checks\" : \[\]}"
+    } else {
+      report_line "No paths found."
+    }
   } else {
```
jjcherry56, line 432: This looks a bit out of place.
I suggest removing the tcl function report_path_ends and making it a swig'ed function
that pushes the prints into ReportPath.cc

akashlevy:
I saw that there was an existing function `ReportPath::reportPathEnds` that was unused (and had some minor issues). I fixed that function and added SWIG interfaces for it. Let me know if this looks ok now or if anything else needs adjustment

akashlevy:
Also, these major changes to path reporting certainly need to be tested on your internal regression. If there are any cases that are failing, please share them with me (either here or via email)

search/Search.tcl
akashlevy: Done

jjcherry56:
The changes are pretty confined to json reporting so it shouldn't be a big deal.
I "always" test on my side before merging PRs.

test/report_json2.tcl
```diff
@@ -0,0 +1,6 @@
+# report_checks -format json with no paths
+read_liberty ../examples/sky130hd_tt.lib.gz
+read_verilog verilog_attribute.v
+link_design counter
+create_clock -name clk -period 10
+report_checks -path_group clk -format json >> results/report_json2.log
```
jjcherry56, line 6: no point in doing redirection here. reporting to the console ends up in the log already. this just breaks the private regressions that use a different directory. I fixed it but for future reference.

akashlevy: Sorry about that!

jjcherry56:
Some other issues to follow up on (I'd rather you dealt with them):
The path_ends vector is leaked (run regression -valgrind report_json1).
This
report_path_ends(PathEndSeq *ends)
  ReportPathFormat path_format = sta->reportPath()->pathFormat();
  if (path_format == ReportPathFormat::json || (ends && ends->size() > 0))
    sta->reportPathEnds(ends);
  else
    report->reportLine("No paths found.");

should be moved down to ReportPath::reportJsonHeader() where there already is an 'if'.

akashlevy:
Thanks, I will deal with these soon

PR #138, opened 2024-12-06
Cache clock domains in check slew limits

SunAnICB:
------

### Background

In a design with 200k components and processing 8 corners at the same time, I encountered a performance problem.

Specifically, **Sta::checkSlewLimits** was very slow in this case. The problem can be located in the call to **clockDomains** in **sta::CheckSlewLimits::findLimit**, which seems to take time **quadratic** to the number of corners.

When **CheckSlewLimits::checkSlewLimits** is using **nullptr** as parameter passed to corner, the inner function call **CheckSlewLimits::checkSlew** will do **CheckSlewLimits::checkSlew1** for all corners. **CheckSlewLimits::checkSlew1** will call **CheckSlewLimits::findLimit** which will call **CheckSlewLimits::clockDomains**. **CheckSlewLimits::clockDomains** will do **VertexPathIterator path_iter(const_cast<Vertex*>(vertex), sta_);** **VertexPathIterator::findNext** will call **arrival_iter_.hasNext()**.

------

### This PR makes the following attempts

- Add a cache for **clockDomains** to **sta::CheckSlewLimits**

  This cache will be cleared each time the **sta::CheckSlewLimits** public function is called.

- Add an option to stop clearing all caches automatically and only clear the corresponding vertex when the pin is disconnected

  ```c++
  sta::CheckSlewLimits::setClockDomainsCanReset(true/false);
  ```

  **The above method can prevent the cache from being completely cleared. However, it is necessary to ensure that the changes during the process will not modify the existing vertex's clock.**

------

Before the change, Sta::checkSlewLimits took about 48 minutes for this case, and about 6 minutes after the change. If the second option is enabled, it will be even faster.

In addition, before this change, I tried 22 corners in the same design, and Sta::checkSlewLimits took about 10 hours plus.

------

### DIFF
--- a/search/CheckSlewLimits.cc
+++ b/search/CheckSlewLimits.cc
@@ -471,4 +471,6 @@ CheckSlewLimits::ClockDomains::clear()
   Super::clear();
 }
 
+thread_local CheckSlewLimits::ClockDomains CheckSlewLimits::clock_domains_;
+
 } // namespace
--- a/search/CheckSlewLimits.cc
+++ b/search/CheckSlewLimits.cc
@@ -185,14 +185,14 @@ CheckSlewLimits::checkSlews1(Vertex *vertex,
 // Return the tightest limit.
 void
 CheckSlewLimits::findLimit0(const Pin *pin,
-                            const Vertex *vertex,
-                            const Corner *corner,
-                            const RiseFall *rf,
-                            const MinMax *min_max,
-                            bool check_clks,
-                            // Return values.
-                            float &limit,
-                            bool &exists) const
+			   const Vertex *vertex,
+                           const Corner *corner,
+			   const RiseFall *rf,
+			   const MinMax *min_max,
+			   bool check_clks,
+			   // Return values.
+			   float &limit,
+			   bool &exists) const
 {
   const Network *network = sta_->network();
   Sdc *sdc = sta_->sdc();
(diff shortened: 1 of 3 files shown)


jjcherry56:
What is the context for the performatnce issue? 
Just calling check_slew_limits or some part of ORFS?
This PR has a number of problems; it will use quite a lot of memory to cache the clocks for every vertex and threads. Maintaining the cache is also problematic.

It seems to me it would be sufficient to just move the clockDomains call outside the corner loop to fix the problem.

jjcherry56:
see c312d5b6 check slews speedup with multiple corners

SunAnICB:
------

It must be admitted that the above commit (c312d5b) solution is indeed more elegant than this PR. And it is undeniable that this PR does have some problems (there is even a crash issue that was not noticed in the previous implementation).

------

My previous scenario is similar to `sta::max_slew_violation_count`.

```c++
// tcl/StaTcl.i
...
size_t
max_slew_violation_count()
{
  cmdLinkedNetwork();
  return Sta::sta()->checkSlewLimits(nullptr, true, nullptr, MinMax::max()).size();
}
...
```

------

From the performance (runtime) perspective, c312d5b is similar to the default mode of this PR. It's not that I haven't considered an implementation like c312d5b. The reason why an optional global cache was finally made was for better performance.

Because in the above scenario, even after c312d5b, clockDomains will still be called repeatedly (of course, it has been greatly improved). The following is my statistics of clockDomains calls when `sta::max_slew_violation_count` is used in this case after c312d5b.

The clockDomains function is called **7.6m** times in total (before c312d5b, this number may be in the order of billions, of course I didn't count it), but the number of different vertex inputs is only **529k** (only one-fourteenth of the total), and the highest number of calls to the same vertex is **345**. In other words, **93%** of the calls to this function (which actually takes a lot of time) can be saved through a global cache, which is the fundamental reason why this PR is implemented in this way.

The optional global cache of my PR can be used as a hack as follows.

```c++
size_t
max_slew_violation_count()
{
  cmdLinkedNetwork();
  sta::CheckSlewLimits::setClockDomainsCanReset(false);
  size_t ret = Sta::sta()->checkSlewLimits(nullptr, true, nullptr, MinMax::max()).size();
  sta::CheckSlewLimits::setClockDomainsCanReset();
  return ret;
}
```

Comparing the initial, c312d5b, current PR (default mode), and the hack based on the current PR (enabling global cache), in this case, the runtime of `sta::max_slew_violation_count` with 8 corners and 22 corners is shown in the following table.

| Solution                                | 8 corners | 22 corners |
| :-------------------------------------- | :-------: | :--------: |
| initial                                 |  0:59:10  |  11:59:41  |
| c312d5b                                 |  0:07:41  |  0:26:27   |
| current PR (default mode)               |  0:07:20  |  0:26:26   |
| hack based on current PR (global cache) |  0:00:47  |  0:02:28   |

> I haven't repeated the test, so the data here is definitely not accurate. It can only reflect the trend.

I don't know if you have a perfect solution for this situation.

------

Overall, the current improvement is great, and the runtime of CheckSlewLimits is already within an acceptable range.

------

PR #139, opened 2024-12-07
Add more allowed characters for RHS of filter expr: `\`, `/`, `$`, `[`, `]`

akashlevy:
**Motivation:** filter name matching is not permissive enough to work with hierarchical full names containing `/`. I realized that a couple more special characters should be allowed on the RHS, so I updated the regex to allow them.

Let me know if I need to add formal test cases for this.

### DIFF
--- a/sdc/Sdc.tcl
+++ b/sdc/Sdc.tcl
@@ -408,7 +408,7 @@ proc current_design { {design ""} } {
 
 # Generic get_* filter.
 proc filter_objs { filter objects filter_function object_type } {
-  set filter_regexp1 {@?([a-zA-Z_]+) *((==|!=|=~|!~) *([0-9a-zA-Z_\*]+))?}
+  set filter_regexp1 {@?([a-zA-Z_]+) *((==|!=|=~|!~) *([0-9a-zA-Z_\\/$\[\]*]+))?}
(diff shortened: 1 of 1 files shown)


jjcherry56:
show me an example. it seems like this could go on forever. it would be better to require the value to be a quoted string or something. It sounds like the user should be using the get_* function instead of get_* <hierarchical_name> instead of get_* -filter full_name=<hierarchical_name>

akashlevy:
I had a customer who wanted to filter `all_registers -data_pins` down further based on the hierarchical name. It seemed apt to add `/` to the allowed RHS characters in that situation and use `sta::filter_objs`.

```
group_path -name AllRet -to [sta::filter_objs {full_name =~ my_module/*} [all_registers -data_pins] sta::filter_pins "pin"]
```

akashlevy:
Any interest in this PR @jjcherry56?

If not, can you let me know, so that I can close this PR? Also, would be good to provide guidance to me on the best way to filter register pins, that way I can relay the guidance to the end user.

jjcherry56:
I would suggest the user write a proc or tcl expression that uses regexp to filter the pins. sta::filter_pins is not exactly a part of the user api.

PR #140, opened 2024-12-09
Apply incremental tolerance to wire slews

povik:
The attached patch saves much of the delay recomputation that happens inside OpenROAD's `repair_timing` command. I've observed any instances touched cause a full recomputation of the output cone.

See the details in the patch -- by applying the incremental delay tolerance to load pin slews most of the recomputation is avoided. In the end this cuts repair_timing runtime on sky130hd/ibex post cts from 170 s to 41 s even if the tolerance is kept at zero (just because it avoids propagation when the slew is exactly the same).

The repair results were identical on the test cases I tried, which gives me some confidence in the patch, but I expect it needs careful review or a rewrite.

Cc @maliberty @QuantamHD @phsauter

### DIFF
--- a/dcalc/GraphDelayCalc.cc
+++ b/dcalc/GraphDelayCalc.cc
@@ -639,6 +639,62 @@ GraphDelayCalc::enqueueTimingChecksEdges(Vertex *vertex)
   }
 }
 
+struct RiseFallSlew {
+  Slew slews[RiseFall::index_count];
+};
+
+static void saveDriverLoadSlews(Graph *graph,
+                                Corners *corners,
+                                Vertex *drvr_vertex,
+                                std::vector<RiseFallSlew> &slews)
+{
+  VertexOutEdgeIterator edge_iter(drvr_vertex, graph);
+  while (edge_iter.hasNext()) {
+    Edge *wire_edge = edge_iter.next();
+    if (wire_edge->isWire()) {
+      Vertex *load_vertex = wire_edge->to(graph);
+      for (auto dcalc_ap : corners->dcalcAnalysisPts()) {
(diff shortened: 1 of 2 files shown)


povik:
> In the end this cuts repair_timing runtime ... even if the tolerance is kept at zero (just because it avoids propagation when the slew is exactly the same).

I've discussed with @precisionmoon and I think in industry terms what this patch does is, even when incremental tolerance is zero, it stops propagating a slew change once it's shielded by the pessimism of a graph-based analysis

jjcherry56:
This begs the question of how the resizer could possibly change a gate without changing the load slews, even after propagating through multiple gates. I would expect to see at least tiny changes in the slews. I have some ideas about why but I gave up investigating for the time being and just accepted they fact.

The following commit implements your insight.
2d11aa9c dcalc use load slew changes to stop propagation

povik:
Thanks, I can confirm your patch has similar impact on runtime.

> question of how the resizer could possibly change a gate without changing the load slews, even after propagating through multiple gates. I would expect to see at least tiny changes in the slews

Isn't this answered by the fact that driver pin slews collect min/max values over multiple paths, so a change on one path doesn't necessarily change the extreme value?

On another topic: I see setting incremental tolerance to a small non-zero value removes further ~30 % of `repair_timing` runtime, but at the same time it degrades the repair results no matter how small the tolerance value I have tested with (10 parts in billion degrades still).

povik:
> setting incremental tolerance to a small non-zero value removes further ~30 % of repair_timing runtime, but at the same time it degrades the repair results no matter how small the tolerance value I have tested with (10 parts in billion degrades still).

I've tracked that down to #148, which I think is the cause for the degradation

PR #144, opened 2024-12-12
Adds error message to prevent segfault

QuantamHD:
make_timing_model segfaults if no liberty library call was run before calling it. This add a helpful error message instead.

### DIFF
--- a/search/Sta.cc
+++ b/search/Sta.cc
@@ -5600,6 +5600,9 @@ Sta::writeTimingModel(const char *lib_name,
                       const char *filename,
                       const Corner *corner)
 {
+  if (network()->defaultLibertyLibrary() == nullptr) {
+    report_->error(2141, "No liberty libraries found.");
+  }
   LibertyLibrary *library = makeTimingModel(lib_name, cell_name, filename,
                                             corner, this);
   writeLiberty(library, filename, this);


search/MakeTimingModel.cc
```diff
@@ -137,6 +138,9 @@ MakeTimingModel::makeLibrary()
 {
   library_ = network_->makeLibertyLibrary(lib_name_, filename_);
   LibertyLibrary *default_lib = network_->defaultLibertyLibrary();
+  if (default_lib == nullptr) {
+    criticalError(2141, "Cannot construct liberty model without first loading a liberty library [read_liberty]");
```
jjcherry56, line 142: criticalError exits OpenSTA. This shouild be an error.
I would prefer the wording to go more like
"No liberty libraries have been read."

The check should be in Sta::writeTiminigModel, as close to the command level as possible instead down in the trenches.

QuantamHD: What is the logger->error equivalent in OpenSTA?

QuantamHD: NVM. Looks like report_->error.

jjcherry56:
I found an analogous check in writePathSpice, so follow it's lead.
report_ and network_ are member vars in Sta but the wording should be sonsistent.

QuantamHD:
Thanks for the pointer to writePathSpice. Everything should be consistent as you suggested

PR #151, opened 2024-12-19
Fix for non-standard latches

akashlevy:
We had a customer face an issue with a Liberty file having a non-standard (read: incorrect) latch expression in a foundry-provided standard cell library. It looked something like this:

```
   latch ("pa pb",IQ,IQN) {
     data_in : "(!C * pa) * !SE + pb * SE";
     enable : "CLK";
     power_down_function : "(!VDD) + (VSS)";
   }
```

It looks like the first expression in the latch group lists all internal pins that are used in the `data_in` expression. (Not sure what useful info that adds, though...)

This produces the error:
```
Critical: /Users/akashlevy/Documents/OpenSTA/liberty/LibertyParser.cc line 390, LibertyStringAttrValue called for float value
```

This PR fixes this issue by detecting when there are three strings. It also includes a test case that fails before and passes after.


test/latch_nonstandard.ok
jjcherry56, line 1: rm empty file

liberty/LibertyReader.cc
```diff
@@ -3970,8 +3970,15 @@ LibertyReader::seqPortNames(LibertyGroup *group,
     else if (i == 1)
       out_inv_name = value->stringValue();
     else if (i == 2) {
-      size = static_cast<int>(value->floatValue());
-      has_size = true;
+      if (value->isFloat()) {
```
jjcherry56, line 3973: I think this would be cleaner if it was rewritten to use switch on the size of the params instead of a for loop. There are really only a few valid cases.

test/regression_vars.tcl
```diff
@@ -132,6 +132,7 @@ record_sta_tests {
   get_objrefs
   get_lib_pins_of_objects
   report_checks_src_attr
+  latch_nonstandard
```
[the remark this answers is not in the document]
akashlevy: sure

liberty/LibertyReader.cc
akashlevy: Agreed. Seems there are three valid cases:
- 2 values: treat as strings for `out_name` and `out_inv_name`
- 3 values:
  - if last value is int, treat first two strings as `out_name` and `out_inv_name`, and last string as size (for `ff_bank` and `latch_bank`)
  - else, ignore first string, treat last two strings as `out_name` and `out_inv_name`

akashlevy:
Amended to address your comments. Please let me know if any further changes should be made

liberty/LibertyReader.cc
jjcherry56: yes

liberty/LibertyReader.cc
```diff
@@ -3959,21 +3959,26 @@ LibertyReader::seqPortNames(LibertyGroup *group,
...
-      out_inv_name = value->stringValue();
-    else if (i == 2) {
-      size = static_cast<int>(value->floatValue());
+  if (group->params()->size() == 2) {
+    out_name = group->firstName();
+    out_inv_name = group->secondName();
+  }
+  else if (group->params()->size() == 3) {
+    LibertyAttrValue *third_value = (*group->params())[2];
+    if (third_value->isFloat()) {
```
jjcherry56, line 3972: comment
// out_port, out_port_inv, bus_size

liberty/LibertyReader.cc
```diff
@@ -3959,21 +3959,26 @@ LibertyReader::seqPortNames(LibertyGroup *group,
...
+      out_name = group->firstName();
+      out_inv_name = group->secondName();
+      size = static_cast<int>(third_value->floatValue());
       has_size = true;
     }
-    i++;
+    else {
+      out_name = group->secondName();
```
jjcherry56, line 3979: comment
// in_port, out_port, out_port_inv, bus_size

akashlevy: There is no `bus_size`, that would be 4 params. Should I add a case for that? Haven't seen it in the wild yet, but might be good to have

jjcherry56: right; I copied too much in the comment.

PR #152, opened 2024-12-20
Fix newline on escaped verilog names

### LINKED ISSUE #146: Escaped verilog names
There is a bug in the verilog file reader. The error is related to escaped verilog names that end with
the newline character '\n' instead of the space character ' '. The situation is correct because in the
"IEEE Standard for Verilog Hardware Description Language" we can see (section 3.7.1):

Escaped identifiers shall start with the backslash character (\) and end with white space (space, tab,
newline). They provide a means of including any of the printable ASCII characters in an identifier (the
decimal values 33 through 126, or 21 through 7E in hexadecimal).

In the verilog file "bus_name.v" we have a bus "\aaa.aaa ". The bus bits are connected to the pins of
the and-gate. The problem is the connection to the pin B. The name of the net "\aaa.aaa [1]" connected
to the pin is divided into two lines (the name of the bus without the space ' ', then the newline
character '\n', several spaces and the bit index [1]). The OpenSTA verilog reader decides that this is
a bit of the bus "\aaa.aaa\n " with a newline character at the end of the name, and creates a new bus,
since a bus with that name does not exist.

The situation is not artifitial. The Cadence synthesis tool Genus performs this separation of bit names.
[bug.zip](https://github.com/user-attachments/files/18124950/bug.zip)

akashlevy:
Resolves #146 by stripping newlines from the end of bus names

### DIFF
--- a/network/ParseBus.cc
+++ b/network/ParseBus.cc
@@ -87,6 +87,10 @@ parseBusName(const char *name,
         is_bus = true;
 	size_t bus_name_len = left - name;
 	bus_name.append(name, bus_name_len);
+	// Remove trailing newlines from bus name.
+	size_t pos;
+	while (((pos=bus_name.find('\n')) != string::npos))
+	  bus_name.erase(pos, 1);
 	// Simple bus subscript.
 	index = atoi(left + 1);
       }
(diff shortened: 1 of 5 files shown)


jjcherry56:
The bug has nothing to do with parsing busses; there could be a newline at the end of any escaped identifier. see
d8a5960 issue146 verilog escaped nanes end in \n

akashlevy:
Thanks, that seems more general for sure

PR #154, opened 2024-12-20
Fix worst slack queue corruption

povik:
Take the following sequence of events:

 * `WorstSlack` is created but there's no related WNS query so the queue is left uninitialized
 * network is edited, the `updateWorstSlack` call inserts endpoints into an uninitialized queue
 * WNS is queried, `WorstSlack` looks into a non-empty queue which has never been initialized and looks up an incorrect WNS value

The patch fixes the above by making sure we don't corrupt the queue in step 2. This bug became visible on a large ORFS test case after 2d11aa9. Thanks to @eder-matheus for initial investigation of the symptoms.


jjcherry56:
took me a while to generate a regression test for this

PR #157, opened 2024-12-25
Add error/warning suppression with `suppress_msg` and `unsuppress_msg` commands

### LINKED ISSUE #147: Way to suppress warnings/errors with command
Would there be interest in having a way to suppress specific warnings/errors with a command? Something like:
```
suppress_msg 1714
```

This would provide a happy medium between displaying all errors/warnings (even expected ones) and using `sta_continue_on_error`. I'm happy to build and upstream this if there is interest.

akashlevy:
Resolves #147.

**List of work:**
- [x] `doc/OpenSTA.odt`, `doc/OpenSTA.pdf`, `doc/ChangeLog.txt`: Add `suppress_msg`/`unsuppress_msg` commands to docs
- [x] `include/sta/Report.hh`, `util/Report.cc`: Add error/warning suppression/unsuppression
- [x] `include/sta/Error.hh`, `util/Error.cc`: Add `suppressed_` private variable + constructor/getter
- [x] `tcl/Exception.i`: Check if `ExceptionMsg` is suppressed before printing to TCL console
- [x] `tcl/Util.tcl`, `util/Util.i`: Add TCL bindings
- [x] `test/suppress_msg.tcl`, `test/suppress_msg.ok`: Simple test of error/warning suppression/unsuppression
- [x] `test/regression_vars.tcl`: Add `suppress_msg` test to regression


tcl/Util.tcl
```diff
@@ -207,6 +207,28 @@ proc sta_warn_error { msg_id warn_error msg } {
   }
 }
 
+define_cmd_args "suppress_msg" {msg_ids}
+
+proc suppress_msg { args } {
+  parse_key_args "suppress_msg" args keys {} flags {}
```
jjcherry56, line 213: no need for this if there are no keywork or flag args

tcl/Util.tcl
```diff
@@ -207,6 +207,28 @@ proc sta_warn_error { msg_id warn_error msg } {
...
+  check_argc_eq1 "suppress_msg" $args
+  set msg_ids [lindex $args 0]
+  foreach msg_id $msg_ids {
+    suppress_msg_id $msg_id
```
jjcherry56, line 217: check_integer "msg_id" $msg_id

tcl/Util.tcl
```diff
@@ -207,6 +207,28 @@ proc sta_warn_error { msg_id warn_error msg } {
...
+  }
+}
+
+define_cmd_args "unsuppress_msg" {msg_ids}
+
+proc unsuppress_msg { args } {
+  parse_key_args "unsuppress_msg" args keys {} flags {}
```
jjcherry56, line 224: as above

util/Report.cc
```diff
@@ -274,10 +306,14 @@ Report::vfileError(int /* id */,
...
 
 void
-Report::critical(int /* id */,
+Report::critical(int id,
                  const char *fmt,
                  ...)
 {
+  // Skip suppressed messages.
+  if (suppressed_msg_ids_.find(id) != suppressed_msg_ids_.end())
+    return;
```
jjcherry56, line 315: You cannot ignore a critical error. This is a condition that means all hope is lost.

util/Report.cc
```diff
@@ -221,9 +237,13 @@ Report::vfileWarn(int /* id */,
 ////////////////////////////////////////////////////////////////
 
 void
-Report::error(int /* id */,
+Report::error(int id,
               const char *fmt, ...)
 {
+  // Skip suppressed messages.
+  if (suppressed_msg_ids_.find(id) != suppressed_msg_ids_.end())
+    return;
```
jjcherry56, line 245: This isn't so simple for errors.
The reason errors throw an exception is because the command will likely seg fault without the check. That is the distinction between warning and error.

tcl/Util.tcl
akashlevy: Removed

tcl/Util.tcl
akashlevy: Done

tcl/Util.tcl
akashlevy: Done

util/Report.cc
akashlevy: Removed suppression

util/Report.cc
akashlevy: Ok. My proposed new solution is to add a private `suppressed` variable to `ExceptionMsg` and check that when the exception is caught. Thoughts?

tcl/Exception.i
```diff
@@ -20,6 +20,13 @@
     fprintf(stderr, "Error: out of memory.\n");
     exit(1);
   }
+  catch (ExceptionMsg &excp) {
+    if (!excp.suppressed()) {
+      Tcl_ResetResult(interp);
+      Tcl_AppendResult(interp, "Error: ", excp.what(), nullptr);
+      return TCL_ERROR;
```
jjcherry56, line 27: The whole point of errors are to throw out of the command because it literally means the 
command cannot continue.
You can suppress the mssage, but not the throw.

include/sta/Error.hh
```diff
@@ -34,11 +34,14 @@ public:
 class ExceptionMsg : public Exception
 {
 public:
-  ExceptionMsg(const char *msg);
+  ExceptionMsg(const char *msg,
+  	       const bool suppressed = false);
```
jjcherry56, line 38: If the suppressed arg is optional it is the same as the first constructor.
I personnally don't want to see optional arg values or initialization in the class.
Restrict initialization to the constructor so the behavior is obvious.

jjcherry56:
Please write up something in doc/ChangeLog.txt

akashlevy:
"The suppress_msg and unsuppress_msg commands allow suppression/unsuppression of
warnings/errors by code."

include/sta/Error.hh
akashlevy: > If the suppressed arg is optional it is the same as the first constructor.
There is just one constructor, I think you may have misread the diff.

I removed initialization of `suppressed_` in the class and restricted it to the constructor.

tcl/Exception.i
akashlevy: Sorry, fixed

doc/ChangeLog.txt
```diff
@@ -31,11 +31,14 @@ is now supported by the the read_saif command.
...
 The report_checks -group_count option has been renamed to -group_path_count.
-The report_checks -endpoing_count option has been renamed to -endpoint_path_count.
+The report_checks -endpoint_count option has been renamed to -endpoint_path_count.
 
-The report_checks -field hierarchical_pins field reports hierarical pins between
+The report_checks -field hierarchical_pins field reports hierarchical pins between
 a driver and a load in the path report.
 
+The suppress_msg and unsuppress_msg commands allow suppression/unsuppression of
+warnings/errors by code.
```
jjcherry56, line 40: explain where to find the code since it is not printed

akashlevy: "Message codes can be found in doc/messages.txt." Added similar language to the docs.

akashlevy:
Synced up the branches. @jjcherry56 anything left to fix?

util/Report.cc
```diff
@@ -169,10 +169,14 @@ Report::printBufferLine()
...
 
 void
-Report::warn(int /* id */,
+Report::warn(int id,
              const char *fmt,
              ...)
 {
+  // Skip suppressed messages.
+  if (suppressed_msg_ids_.find(id) != suppressed_msg_ids_.end())
+    return;
```
jjcherry56, line 178: I am no fan of returns like this. ]

test/suppress_msg.tcl
```diff
@@ -0,0 +1,16 @@
+# suppress and unsuppress message ids
+
+set sta_continue_on_error 1
+
+::sta::report_warn 1234 "EXPECTED: This is a warning message"
```
jjcherry56, line 5: leading :: is unnecessary

akashlevy: Removed

util/Report.cc
akashlevy: See now?

jjcherry56:
I think this needs more through testing.
I would use catch instead on continue_on_error.
I am not convinced that the changes to Report::error and ExceptionMsg are necessary; it
seems like the behavior is the same for a suppressed error (which I think is correct).

proc cmd {} {
  sta::sta_warn 1 "cmd error"
  sta::sta_error 2 "cmd error"
  puts "after error"
}

catch { cmd } error
puts $error

suppress_msg 1
suppress_msg 2
catch { cmd } error
puts $error

akashlevy:
I see the whole point of error message suppression as affecting the message shown when using `sta_continue_on_error`. In your example above, it's doing nothing.

I propose we either remove error message suppression altogether or have it suppress the message when `sta_continue_on_error` is used (current behavior).

jjcherry56:
I don't see it suppressing calls 3 and 4 wth sta_continue_on_error

[suppress.txt](https://github.com/user-attachments/files/18311575/suppress.txt)

I think the behavior for errors is far from obvious, which is why I let this sit for so long.
What about separating error handling from warings with suppress_warning and
then think some more on suppress_error?

akashlevy:
Ah, I see the issue.

The issue with your test is that you are using `sta::sta_error` rather than `sta::report_error`. `sta::sta_error` doesn't actually raise an exception if you look at the TCL for it. This means that if an error originates from the TCL side using `sta::sta_error`, it won't be suppressed.

However, if the error originates from C++ (which uses `report->error`) it will be suppressed properly. You can see that if you use replace `sta::sta_error` with `sta::report_error` (which calls `report->error`), your example works as expected, yielding:
```
call 1
Warning: cmd error
Error: cmd error
call 2
caught 
call 3
call 4
```

One thing I could do is maintain a list of suppressed messages on the TCL side as well, and check against those in `sta::sta_error`. Does this seem like a valid solution?

akashlevy:
@jjcherry56 see my new solution for dealing with errors on the TCL side. I updated the test case to be yours

jjcherry56:
sta::report_error is not used (check the tcl files).

akashlevy:
Because I took it from your example. Do errors using `sta::report_error` need to be handled? If so, you are asking me to put back the `ExceptionMsg` changes, which is the best way I can see to do that

jjcherry56:
No, there is no point in handling sta::report_error. It should just be deleted because it is a swig function that does not support a tcl function like the others.

jjcherry56:
Actually you can use sta::report_error to test calls to report->error from c++ in tcl.

akashlevy:
@jjcherry56 Ok, I added back in C++ side error suppression and now have a test case that checks that errors generated on both C++/TCL side can be suppressed properly. Let me know if anything else needs fixing

tcl/Util.tcl
```diff
@@ -207,6 +213,31 @@ proc sta_warn_error { msg_id warn_error msg } {
...
+
+define_cmd_args "unsuppress_msg" {msg_ids}
+
+proc unsuppress_msg { args } {
+  check_argc_eq1 "unsuppress_msg" $args
+  set msg_ids [lindex $args 0]
+  foreach msg_id $msg_ids {
+    check_integer "msg_id" $msg_id
+    unsuppress_msg_id $msg_id
+    lappend suppressed_msgs $msg_id
```
jjcherry56, line 237: this needs to remove msg_id, not append it. but suppressed_msgs should go away...

tcl/Util.tcl
```diff
@@ -192,10 +194,14 @@ proc sta_warn { msg_id msg } {
...
   variable sdc_file
   variable sdc_line
-  if { [info exists sdc_file] } {
-    error "Error: [file tail $sdc_file] line $sdc_line, $msg"
-  } else {
-    error "Error: $msg"
+  variable suppressed_msgs
+
+  if { ! [info exists suppressed_msgs] || ! $msg_id in $suppressed_msgs } {
+    if { [info exists sdc_file] } {
```
jjcherry56, line 200: use a function defined in Util.i to query instead of maintaining a separate list

akashlevy:
See now with requested fixes

akashlevy:
I added a test for unsuppression as well now

tcl/Util.tcl
```diff
@@ -207,6 +210,29 @@ proc sta_warn_error { msg_id warn_error msg } {
   }
 }
 
+define_cmd_args "suppress_msg" {msg_ids}
+
+proc suppress_msg { args } {
+  variable suppressed_msgs
+  check_argc_eq1 "suppress_msg" $args
```
jjcherry56, line 217: Instead of taking a list of ids as one arg I think it should just use the args.
So
suppress_msg 1 2 3
instead of 
suppress_msg {1 2 3}

tcl/Util.tcl
```diff
@@ -207,6 +210,29 @@ proc sta_warn_error { msg_id warn_error msg } {
   }
 }
 
+define_cmd_args "suppress_msg" {msg_ids}
```
jjcherry56, line 213: msg_ids...

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+  sta::sta_warn 1 "cmd error"
+  sta::sta_error 2 "cmd error"
+  puts "after error"
+}
+
+# Run report_warn/report_error to test C++ side suppression
+proc report_cmd { msg } {
+  puts "report_cmd call $msg"
+  sta::report_warn 1 "cmd error"
+  sta::report_error 2 "cmd error"
```
jjcherry56, line 15: "cmd warn"

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
+# suppress and unsuppress message ids
+
+# Run sta_warn/sta_error to test TCL side suppression
+proc sta_cmd { msg } {
+  puts "sta_cmd call $msg"
+  sta::sta_warn 1 "cmd error"
```
jjcherry56, line 6: "cmd warn"

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+  sta::report_error 2 "cmd error"
+  puts "after error"
+}
+
+# Ensure that TCL side messages are displayed as usual
+catch { sta_cmd 1 } error
+puts $error
+
+# Ensure that C++ side messages are displayed as usual
+catch { report_cmd 1 } error
```
jjcherry56, line 24: report_cmd 2

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+puts $error
+
+# Suppress messages
+suppress_msg 1
+suppress_msg 2
+
+# Ensure that TCL side messages are suppressed
+catch { sta_cmd 2 } error
```
jjcherry56, line 32: report_cmd 3

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+puts "caught $error"
+
+# Ensure that C++ side messages are suppressed
+catch { report_cmd 2 } error
```
jjcherry56, line 36: report_cmd 4

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+puts "caught $error"
+
+# Continue on error to avoid having to catch
+set sta_continue_on_error 1
+
+# Ensure that TCL side messages are suppressed
+# TCL side will make it to "after error"
+sta_cmd 3
```
jjcherry56, line 44: report_cmd 5

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+sta_cmd 4
```
jjcherry56, line 45: report_cmd 6

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+
+# Ensure that C++ side messages are suppressed
+# C++ will not make it to "after error" as the whole cmd is cancelled
+report_cmd 3
```
jjcherry56, line 49: report_cmd 7

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+report_cmd 4
```
jjcherry56, line 50: report_cmd 8

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+
+# Unsuppress messages
+unsuppress_msg 1
+unsuppress_msg 2
+
+# Ensure that TCL side messages are displayed as usual
+catch { sta_cmd 5 } error
```
jjcherry56, line 57: report_cmd 9

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+puts $error
+
+# Ensure that C++ side messages are displayed as usual
+catch { report_cmd 5 } error
```
jjcherry56, line 61: report_cmd 10

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+  puts "after error"
+}
+
+# Ensure that TCL side messages are displayed as usual
+catch { sta_cmd 1 } error
+puts $error
+
+# Ensure that C++ side messages are displayed as usual
+catch { report_cmd 1 } error
+puts $error
```
jjcherry56, line 25: uts "caught $error"

test/suppress_msg.tcl
```diff
@@ -0,0 +1,62 @@
...
+report_cmd 3
+report_cmd 4
+
+# Unsuppress messages
+unsuppress_msg 1
+unsuppress_msg 2
+
+# Ensure that TCL side messages are displayed as usual
+catch { sta_cmd 5 } error
+puts $error
```
jjcherry56, line 58: uts "caught $error"

akashlevy:
@jjcherry56 See requested fixes

test/suppress_msg.tcl
```diff
@@ -0,0 +1,60 @@
+# suppress and unsuppress message ids
+
+# Run sta_warn/sta_error to test TCL side suppression
+proc sta_cmd { msg } {
+  puts "sta_cmd call $msg"
```
jjcherry56, line 5: remove the puts and add $msg to the warn/error calls here and in the next proc

akashlevy:
Good now?

test/suppress_msg.tcl
```diff
@@ -0,0 +1,58 @@
...
+# Run report_warn/report_error to test C++ side suppression
+proc report_cmd { msg } {
+  sta::report_warn 1 "cmd warn $msg"
+  sta::report_error 2 "cmd error $msg"
+  puts "after error"
+}
+
+# Ensure that TCL side messages are displayed as usual
+catch { sta_cmd 1 } error
+puts $error
```
jjcherry56, line 19: puts "caught $error"

test/suppress_msg.tcl
```diff
@@ -0,0 +1,58 @@
...
+
+# Ensure that C++ side messages are displayed as usual
+catch { report_cmd 2 } error
+puts $error
```
jjcherry56, line 23: puts "caught $error"

test/suppress_msg.tcl
```diff
@@ -0,0 +1,58 @@
...
+catch { report_cmd 4 } error
+puts "caught $error"
+
+# Continue on error to avoid having to catch
+set sta_continue_on_error 1
+
+# Ensure that TCL side messages are suppressed
+# TCL side will make it to "after error"
+sta_cmd 5
+sta_cmd 6
```
jjcherry56, line 42: I don't understand why there are 2 calls here when the "after error" indicates it continuted.

test/suppress_msg.tcl
```diff
@@ -0,0 +1,58 @@
...
+
+# Ensure that C++ side messages are suppressed
+# C++ will not make it to "after error" as the whole cmd is cancelled
+report_cmd 7
+report_cmd 8
```
jjcherry56, line 47: I don't understand why there are 2 calls here when the "after error" indicates it continuted.

test/suppress_msg.tcl
akashlevy: Switched to single call

test/suppress_msg.tcl
akashlevy: Switched to single call

akashlevy:
In my opinion, your request to remove the initial `puts` from `sta_cmd` and `report_cmd` procs makes it much more difficult to understand which call is running when looking at the log.

jjcherry56:
I think this is ready but I see conflicts in test/regression_vars.tcl and OpenSTA.odt
I have held back on making changes to OpenSTA.odt until this is merged

akashlevy:
Really? It says no conflicts on my side
image

akashlevy:
I resolved a few conflicts a couple days ago, make sure the branch is up to date?

PR #171, opened 2025-01-13
Optimize class ExceptionStateSet to improve both runtime and memory usage.

ZhangChangICB:
---
motivation && reason
---
Under the original ExceptionStateSet architecture, we observed that during the process of exception propagation, **the STA would attempt to create an ExceptionStateSet for each specific Vertex and Tag**, even if it was entirely unrelated to that point and could have been completely copied from the ExceptionStateSet of the previous stage. Like following:
```
if (states == nullptr)
  states = new ExceptionStateSet();
states->insert(state);
```

This meant that a large number of identical ExceptionStateSet instances were created, even though their contents were exactly the same, leading to **unnecessary memory overhead**. 

Furthermore, under this architecture, every time a Tag Match was performed, the STA needed to repeatedly compare the elements within the ExceptionStateSet. Additionally, when calculating the Tag Hash, it had to repeatedly iterate over and compute the hash values of each Exception in the ExceptionStateSet. This resulted in **significant additional runtime overhead**.

---
action && improvement
---
We refactored the ExceptionStateSet structure and observed over a 5% runtime improvement through testing with the pprof tool on our test design, along with minor memory optimizations. The changes we made include the following:

1. Ensured that ExceptionStateSet instances with identical contents are treated as the same element.
2. With the above adjustment, Tag Match/Exception Match operations can now match pointers instead of comparing their actual contents.
3. Cached the hash of the ExceptionStateSet, eliminating the need to recalculate it every time a Tag is created.
4. Replaced the use of OwnState for memory management with a reference counting (refCount) approach.

---
pprof data && evidence
---

> Under design with about 150k instances, 4 corners, dozens of exceptions && clocks.
> Read design and find requireds.

- Before this pull request:
[before_this_pr.pdf](https://github.com/user-attachments/files/18393820/before_this_pr.pdf)

- After this pull request:
[after_this_pr.pdf](https://github.com/user-attachments/files/18393825/after_this_pr.pdf)
_**You can see the sample time percentage of function "findTag" improved from 12.3% to 5.9%.**_
---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,7 @@ set(STA_SOURCE
   parasitics/SpefReader.cc
   parasitics/SpefReaderPvt.hh
   
+  sdc/SdcClass.cc
   sdc/Clock.cc
   sdc/ClockGatingCheck.cc
   sdc/ClockGroups.cc
--- a/include/sta/Sdc.hh
+++ b/include/sta/Sdc.hh
@@ -993,27 +993,27 @@ public:
 			   const Clock *clk,
 			   const RiseFall *clk_rf,
 			   const MinMax *min_max,
-			   ExceptionStateSet *&states) const;
+			   ExceptionStates &states) const;
   bool exceptionFromStates(const Pin *pin,
 			   const RiseFall *rf,
 			   const Clock *clk,
 			   const RiseFall *clk_rf,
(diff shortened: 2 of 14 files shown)


jjcherry56:
This changes too much code for too little benefit

PR #174, opened 2025-01-13
Verilog reader enhancement to support swapping to uninstantiated modules

precisionmoon:
1) mark black box cells during linking to exclude them from swapping candidates
2) defer module deletion to VerilogReader destructor
3) use actual module name name when linking unused modules as top cell

### DIFF
--- a/verilog/VerilogReader.cc
+++ b/verilog/VerilogReader.cc
@@ -28,6 +28,8 @@
 #include "VerilogNamespace.hh"
 #include "StringUtil.hh"
 #include "verilog/VerilogReaderPvt.hh"
+#include "ConcreteNetwork.hh"
+#include "NetworkClass.hh"
 
 extern int
 VerilogParse_parse();
@@ -48,6 +50,7 @@ Instance *
 linkVerilogNetwork(const char *top_cell_name,
 		   bool make_black_boxes,
 		   Report *report,
+                   bool use_top_cell_name,
 		   NetworkReader *network);
 
 bool
@@ -1785,9 +1788,11 @@ Instance *
 linkVerilogNetwork(const char *top_cell_name,
 		   bool make_black_boxes,
 		   Report *report,
+                   bool use_top_cell_name,
 		   NetworkReader *)
 {
-  return verilog_reader->linkNetwork(top_cell_name, make_black_boxes, report);
+  return verilog_reader->linkNetwork(top_cell_name, make_black_boxes, report,
+                                     use_top_cell_name);
 }
 
 // Verilog net name to network net map.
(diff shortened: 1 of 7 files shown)


verilog/VerilogReader.cc
```diff
@@ -1815,14 +1820,17 @@ class VerilogBindingTbl
...
+			   Report *report,
+                           bool use_top_cell_name)
 {
   if (library_) {
     Cell *top_cell = network_->findCell(library_, top_cell_name);
     VerilogModule *module = this->module(top_cell);
     if (module) {
       // Seed the recursion for expansion with the top level instance.
-      Instance *top_instance = network_->makeInstance(top_cell, "", nullptr);
+      // When uninstantiated modules are linked, we need actual top cell name
```
jjcherry56, line 1831: No, WE do not need the actual top cell name. YOU need the actual top cell name. Don't pretend you are the only user of this code.

I changed this to always set the instance name, getting rid of all the api changes in the caller.

verilog/VerilogReader.cc
```diff
@@ -1915,6 +1925,10 @@ VerilogReader::makeModuleInstNetwork(VerilogModuleInst *mod_inst,
     string inst_vname = verilogName(mod_inst);
     if (make_black_boxes) {
       cell = makeBlackBox(mod_inst, parent_module);
+      // Mark black box cells such that these can be
+      // excluded for candidates for uninstantiated module swap
+      ConcreteCell* ccell = reinterpret_cast<ConcreteCell*>(cell);
+      ccell->setAttribute("black_box", "true");
```
jjcherry56, line 1931: This is a gross violation of the apis. This file should not be including ConcreteNetwork.hh, and this should be using Network::setAttribute instead. But I really believe all of the attribute setting should be moved out of OpenSTA and done on the OR side (write a predicate to detect the black box).

verilog/VerilogReader.cc
```diff
@@ -1843,7 +1851,9 @@ VerilogReader::linkNetwork(const char *top_cell_name,
       }
       makeModuleInstBody(module, top_instance, &bindings, make_black_boxes);
       bool errors = reportLinkErrors(report);
-      deleteModules();
+      // To support swapping to uninstantiated modules, defer module deletion
+      // to VerilogReader destructor
+      // deleteModules();
```
jjcherry56, line 1856: IT NEVER makes sense to comment out code and then write a comment about why it is commented out. The comment may make sense the but the dead code does not. The modules are deleted to reduce the peak memory (for flattened netlists the memory for the network is doubled without this). This may mean OR needs to define a different link function that passes in an arg that suppresses deletion. But this is not the solution.

verilog/VerilogReader.cc
precisionmoon: OK.  Thanks.

verilog/VerilogReader.cc
precisionmoon: OK, I see the difference in link behavior between OR and OpenSTA.    OR deletes VerilogReader immediately after linking.  Can Sta::linkDesign delete modules after linking?

verilog/VerilogReader.cc
precisionmoon: OK.  Let me think about doing this outside OpenSTA.   Can the rest of the changes be merged then?

precisionmoon:
Closing this PR in light of recent changes

PR #181, opened 2025-01-16
Report json fixes

akashlevy:
- Move "No paths found." logic to `ReportPath::reportPathEnds` where it belongs
- Delete `ends` to fix memory leak


search/ReportPath.cc
```diff
@@ -335,7 +335,7 @@ void
 ReportPath::reportPathEnds(PathEndSeq *ends)
 {
   reportPathEndHeader();
-  if (ends) {
+  if (ends && ends->size()) {
```
jjcherry56, line 338: !ends->empty()

PR #182, opened 2025-01-16
Migrate static helper functions in `StaTclTypes.i` to non-static functions in `TclTypeHelpers.cc`/`.hh`

akashlevy:
See #117


CMakeLists.txt
```diff
@@ -225,6 +225,8 @@ set(STA_SOURCE
   
   verilog/VerilogReader.cc
   verilog/VerilogWriter.cc
+
```
jjcherry56, line 228: this should be alphabetical in the list (power/ is wrong too)

CMakeLists.txt
```diff
@@ -225,6 +225,8 @@ set(STA_SOURCE
...
+  tcl/StaTclTypeHelpers.cc
```
jjcherry56, line 229: this should be alphabetical in the list (power/ is wrong too)

the same is pretty long; how about tightening it up to TclTypeHelpers?

akashlevy: Sure

PR #193, opened 2025-01-24
Enhanced Verilog reader to support swapping to uninstantiated modules

precisionmoon:
1) Removed module deletion from linkNetwork
2) Added module deletion to postLinkNetwork

### DIFF
--- a/include/sta/ConcreteNetwork.hh
+++ b/include/sta/ConcreteNetwork.hh
@@ -68,6 +68,7 @@ public:
   bool linkNetwork(const char *top_cell_name,
                    bool make_black_boxes,
                    Report *report) override;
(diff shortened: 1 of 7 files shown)


jjcherry56:
This is way too much mucking around.
I suggest you add a delete_modules arg to VerilogReader::linkNetwork, make Sta::linkDesign virutal and then make dbSta::linkDesign pass in false.

precisionmoon:
> This is way too much mucking around. I suggest you add a delete_modules arg to VerilogReader::linkNetwork, make Sta::linkDesign virutal and then make dbSta::linkDesign pass in false.

OK.  I'll close this PR and open a new one with the suggested changes.

PR #194, opened 2025-01-26
DO NOT MERGE testing faster map for vcd reader

oharboe:
Checking in the debugger for a large .vcd file(320gByte), I see that all the time is spent in map::find().

Quick test for parsing large .vcd files, ca. 8x faster with `absl::flat_hash_map`

Tried std::unordered_map, but I could not tell a difference in performance.

Running a large test now that took ca. 14 hours, it runs in 109 minutes, so ca. 8x faster.

boost unrdered map, 198 minutes

Even with this improvement in in map::find(), most of the time does appear to be spend in the find() function, so perhaps there is an even more suitable map implementation for large .vcd files.

Perhaps a custom hashing function for std::map() would do the trick?

trying boost unordered flat map available in boost 1.83, looks like it will match absail unorderef flat map 104 minutes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ doc/._Sta.docx
 test/results
 # ngspice turd
 test/b3v3_1check.log
+*.a
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -516,6 +516,7 @@ target_sources(OpenSTA
 )
 
 target_link_libraries(OpenSTA
+  PUBLIC
   Eigen3::Eigen
   ${TCL_LIBRARY}
   ${CMAKE_THREAD_LIBS_INIT}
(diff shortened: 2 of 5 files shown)


QuantamHD:
@jjcherry56 is unlikely to take a dep on abseil or boost. Those are really big deps when you only really need a single thing. He might be willing to take something like robinhood hash which is just a single file that could be added to util. https://github.com/martinus/robin-hood-hashing/blob/master/src/include/robin_hood.h, but I don't know either.

oharboe:
@maliberty @QuantamHD So, this is a solved problem, we just don't want the dependencies. Ideas?

maliberty:
I do find it surprising that map and unordered_map would be equal in performance.  Are you certain about that result?

oharboe:
> I do find it surprising that map and unordered_map would be equal in performance. Are you certain about that result?

There might have been a difference, but it didn't "move the needle" compared to unordered_flat_map. c++-23 has an unordered_flat_map, but I didn't try that, not available out of the box in Ubuntu 24.04.

oharboe:
@QuantamHD @maliberty Wrote a bit of code that converted an autogenerated .vcd id to std::vector index. Measuring now... Looks like it is on par with boost:unordered_flat_map. It is a slight complication, needs some more work, but works well for verilator generated .vcd files and probably other large .vcd files as well. Measuring now.

PR #200, opened 2025-01-30
made module deletion optional in linkNetwork

precisionmoon:


jjcherry56:
WTF, it doesn't even compile with these changes. Now I get to clean up the mess.

Issue #204, opened 2025-02-03
CentOS 7: invalid command name "try" after `source set info script`


akashlevy:
See this run: https://github.com/Silimate/OpenSTA/actions/runs/13119278783/job/36601113721

Basically, `try`-`finally` is not supported in TCL 8.5, which is the latest version available from the repos for CentOS 7. A few options for CentOS/RHEL 7 compatibility:
1. Use an alternate to `try`-`finally` for https://github.com/parallaxsw/OpenSTA/commit/dbd95cce7d00866172d9a92b190e7eac5b8c3042 (e.g., `catch`)
2. Use a forward-compatible `try`: https://wiki.tcl-lang.org/page/Forward%2Dcompatible+try+and+throw
3. Build and install TCL 8.6+ from source inside CentOS 7 `Dockerfile`

Let me know which is preferred. I can make a PR for (2) or (3) if needed

akashlevy:
Here's how we solved it on our side using option 2: https://github.com/Silimate/OpenSTA/commit/144fe107934c93b01dfafd344768a391ae83e230

jjcherry56:
I think it would be sufficient to rewrite it to use catch

PR #205, opened 2025-02-04
rename sta::source to sta::include to allow use of tcl source

### LINKED ISSUE #202: reimplemented `source` executes all commands at top-level, breaking typical tcl codebases
The reimplementation of `source` runs all commands at the top-level namespace:

https://github.com/parallaxsw/OpenSTA/blob/41d1d1bb7b045118705a1cad64d68a0d78343f52/sdc/Sdc.tcl#L109

This causes incompatibilities with typical Tcl codebases, including but not limited to the Tcl standard library, which rely on scoped `source` statements:

https://github.com/tcltk/tcl/blob/874e4fe4264a40c00c4db5115afba9600f9f368d/library/package.tcl#L491

A minimal reproducer of this is as follows:

sourceable.tcl:
```
puts "Hi, $name!"
```

main.tcl
```
proc say_name {name} {
    source ./sourceable.tcl
}

say_name donn
```

```
$ tclsh main.tcl
Hi, donn!
$ sta -exit main.tcl
Error: sourceable.tcl, 1 can't read "name": no such variable
```

---

A simple solution would be to change the uplevel to `2` (to account for `source`, `source_`) and uplevel the source invocation in `read_sdc` itself to `\#0`:

```diff
diff --git a/sdc/Sdc.tcl b/sdc/Sdc.tcl
index d2512ba8..9e9adf14 100644
--- a/sdc/Sdc.tcl
+++ b/sdc/Sdc.tcl
@@ -41,8 +41,12 @@ proc_redirect read_sdc {
 
   check_argc_eq1 "read_sdc" $args
   set echo [info exists flags(-echo)]
+  set echo_flag ""
+  if { $echo } {
+    set echo_flag -echo
+  }
   set filename [file nativename [lindex $args 0]]
-  source_ $filename $echo 0
+  uplevel \#0 "source $echo_flag $filename"
 }
 
 ################################################################
@@ -106,7 +110,7 @@ proc source_ { filename echo verbose } {
       if { [string index $line end] != "\\" \
 	     && [info complete $cmd] } {
 	set error {}
-	set error_code [catch {uplevel \#0 $cmd} result]
+	set error_code [catch {uplevel 2 $cmd} result]
 	# cmd consumed
 	set cmd ""
 	# Flush results printed outside tcl to stdout/stderr.

```

donn:
- create new global stack, ::_sta_file_stack, to keep track of evaluated files (include and source)
- rename sta::source to sta::include, allowing the original tcl source (which behaves differently, including inheriting the caller's scope) to be used
- create new thin wrapper for tcl source as sta::source - all it does is push to and pop from ::_sta_file_stack
- update sta_error, sta_warn to inspect the stack and alter behavior based on whether the current file is sourced or included 

Original PR body follows:

---

- add new global variable, ::sta_scoped_source that if set to 1, causes all sourced files to be evaluated at the scope of the calling function instead of the top level (default 0). read_sdc will always be evaluated at the top-level, however.
- ignore early returns when sourcing files - tcl return is supposed to work when sourcing files. see https://www.tcl-lang.org/man/tcl8.6/TclCmd/return.htm for more info

---

I have chosen to make it opt-in behavior out of an excess of caution. The `~/.sta` issue could technically be resolved by upleveling `sourceTclFile`, but that may have unforeseen consequences. Fortunately, it's not a huge ask to ask users to set ::sta_scoped_source to 1 before package require statements.

Resolves #202

### DIFF
--- a/app/StaMain.cc
+++ b/app/StaMain.cc
@@ -98,7 +98,7 @@ sourceTclFile(const char *filename,
 	      Tcl_Interp *interp)
 {
   string cmd;
-  stringPrint(cmd, "read_cmds %s%s%s",
+  stringPrint(cmd, "include %s%s%s",
 	      echo ? "-echo " : "",
 	      verbose ? "-verbose " : "",
 	      filename);
--- a/app/StaMain.cc
+++ b/app/StaMain.cc
@@ -90,7 +90,7 @@ findCmdLineKey(int &argc,
   return nullptr;
 }
 
-// Use overridden version of source to echo cmds and results.
+// Use custom source function to echo cmds and results.
 int
 sourceTclFile(const char *filename,
 	      bool echo,
@@ -98,7 +98,7 @@ sourceTclFile(const char *filename,
 	      Tcl_Interp *interp)
 {
   string cmd;
-  stringPrint(cmd, "source %s%s%s",
+  stringPrint(cmd, "read_cmds %s%s%s",
 	      echo ? "-echo " : "",
 	      verbose ? "-verbose " : "",
 	      filename);
--- a/sdc/Sdc.tcl
+++ b/sdc/Sdc.tcl
@@ -45,7 +45,7 @@ proc_redirect read_sdc {
   set prev_filename [info script]
   try {
     info script $filename
-    source_ $filename $echo 0
+    uplevel \#0 "sta::source_ $filename $echo 0"
   } finally {
     info script $prev_filename
   }
@@ -63,6 +63,7 @@ if { ![info exists renamed_source] } {
 }
 
 set ::sta_continue_on_error 0
+set ::sta_scoped_source 0
 
 define_cmd_args "source" \
   {[-echo] [-verbose] filename [> filename] [>> filename]}
@@ -79,7 +80,7 @@ proc_redirect source {
   set prev_filename [info script]
   try {
     info script $filename
-    source_ $filename $echo $verbose
+    uplevel 1 "sta::source_ $filename $echo $verbose"
   } finally {
     info script $prev_filename
   }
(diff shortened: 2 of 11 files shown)


donn:
sorry - found another issue with package require,

akashlevy:
Any chance you can resolve #204 while you are at it? Would greatly appreciate it @donn

donn:
Hi Akash- will try, pun not intended, but I really doubt I can because this whole thing is already enough of a Tcl scoping nightmare. I can't just include the forward-compat code either as it has no license.

akashlevy:
If you can use `catch` instead of `try`, that would be awesome

akashlevy:
This is addressed in #206 so no need to address @donn, thanks anyways

donn:
@jjcherry56 Should be ready now.

jjcherry56:
I am not a fan of this approach.
The redefinition of source dates back 20 years to match the behavior of another tool.
I think it would make more sense to remove the code that redefines source and define a new command "read_cmds" (without the unused -verbose option) based on the existing implementation.

donn:
That... would be my preferred approach too. I just assumed you wouldn't want to break existing code.

donn:
@jjcherry56 So I implemented a solution to this that works as follows:

There's now an `::sta_file_stack`. This stack is composed of tuples of:
- either include or source
- the normalized filename

Both `sta::include` and `sta::source`	add the file currently under processing to the stack.

`sta::warn` and `sta::error` peek the top element of the stack, and if it's an `include`, use the previous behavior.

If it's a `source`, however, the frame stack is inspected to return the line number of the *lowest stack frame in the sourced file*. This is to emulate the behavior of how `include`'s line numbers work as closely as possible. I added a test for this behavior.

jjcherry56:
what I had in mind was NOT redefining source and making include set the info file/line properties like source does so it has the same behavior and modifying sta_error/warn to use the info file/line instead of variables or stack

PR #210, opened 2025-02-06
Messages: Generates messages ids during build and stop in case of duplicate and fix duplicate messages ids

titan73:


titan73:
Update PR not to fail build on duplicate ids and turn errors messages into warnings.

.gitignore
```diff
@@ -30,3 +30,5 @@ doc/._Sta.docx
 test/results
 # ngspice turd
 test/b3v3_1check.log
+
+messages.txt
```
jjcherry56, line 34: doc/messages.txt

doc/messages.txt
jjcherry56, line 1: this should be rm'd from git now that it is auto generated (can't be ignored and in the repo)

PR #215, opened 2025-02-12
All path groups

### LINKED ISSUE #211: No way to enumerate path_groups in Tcl
I can't find a way to list all the path groups created by users in diverse .sdc files. It would be handy to be able to collect these to create automatic reporting.

akashlevy:
Resolves #211 by adding `sta::all_path_groups` command that returns the names of all path groups.

Let me know what tweaks are needed.


sdc/Sdc.i
```diff
@@ -1323,6 +1323,17 @@ filter_timing_arcs(const char *property,
 
 ////////////////////////////////////////////////////////////////
 
+StringSeq
+all_path_groups()
```
jjcherry56, line 1327: how about path_group_names because it does not return the objects

sdc/Sdc.i
```diff
@@ -1323,6 +1323,17 @@ filter_timing_arcs(const char *property,
 
 ////////////////////////////////////////////////////////////////
 
+StringSeq
+path_group_names()
+{
+  StringSeq path_groups;
+  for (auto gp : Sta::sta()->sdc()->groupPaths())
+    path_groups.push_back(gp.first);
```
jjcherry56, line 1331: use c++17 iteration
for (auto const& [name, group] :

PR #218, opened 2025-02-16
Add LibertyCell::leakagePower overload using std::optional

maliberty:
More modern C++ style

### DIFF
--- a/liberty/Liberty.cc
+++ b/liberty/Liberty.cc
@@ -1307,6 +1307,16 @@ LibertyCell::leakagePower(// Return values.
   exists = leakage_power_exists_;
 }
 
+std::optional<float>
+LibertyCell::leakagePower() const
+{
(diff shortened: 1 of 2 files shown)


liberty/Liberty.cc
```diff
@@ -1307,6 +1307,16 @@ LibertyCell::leakagePower(// Return values.
   exists = leakage_power_exists_;
 }
 
+std::optional<float>
+LibertyCell::leakagePower() const
+{
+  std::optional<float> leakage;
+  if (leakage_power_exists_) {
```
hzeller, line 1314: You  don't necessarily need the `leakage` value to assign stuff to and simplify that a bit by directly returning the value:

```c++
{
  if (leakage_power_exists_) {
     return leakgage_power_;
  }
  return std::nullopt;
}
```

hzeller:
Nice!

PR #230, opened 2025-04-07
Put generated StaConfig.hh into build artifacts.

hzeller:
Before, the cmake configure_file() for the StaConfig.hh header, that make only sense for the particular cmake configuration as it contains specific defines for that build configuration, ended up in the original source tree.

With this fix, it is a build artifact inside build/.

Tested, that the final `make install` properly copies the StaConfig.hh to the correct location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,6 @@ cmake-build-debug
 build
 pvt
 
-include/sta/StaConfig.hh
-
 app/sta
 app/libOpenSTA.*
 app/sta.exe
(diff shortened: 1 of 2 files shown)


parallaxsw:
Is there some reason it is in build/include/sta instead of just build or build/include?

hzeller:
I just wanted to keep it as close as possible to the original, and since all includes reference `include/sta/` as their `-I` location that also felt like a natural choice to not confuse people later.

hzeller:
Ready to merge or do you want me to change something ?

PR #232, opened 2025-04-08
Fix use-after-free caused by freeing pointers owned by another object.

Logikable:
TableAxis's values_ are at least sometimes (if not always) owned by LibertyReader, originating from axis_values_. It's better we delete them in LibertyReader.

We add the test `reused_table_axis` to prevent backsliding.


QuantamHD:
@Logikable works for me at Google.

liberty/TableModel.cc
```diff
@@ -1433,7 +1433,6 @@ TableAxis::TableAxis(TableAxisVariable variable,
 
 TableAxis::~TableAxis()
 {
-  delete values_;
 }
```
parallaxsw, line 1436: this leaks the values_ array

liberty/LibertyReader.cc
```diff
@@ -1520,6 +1520,8 @@ LibertyReader::visitIndex(int index,
           prev = value;
         }
       }
+      if (axis_values_[index])
```
parallaxsw, line 1523: delete does not require conditionalization

Logikable:
Thank you.

PR #243, opened 2025-05-12
Fix levelize crash when vertices are enqueued in arrival/required iterators

povik:
Fix for re-levelization crashing if vertices are queued up in `search_->arrival_iter_` or `search_->required_iter_`.

The sequence leading up to the crash is the following:

 * Arrivals are computed, vertices are levelized.
 * The network is edited, some vertices are queued up in the arrival/required iterators. Levelization is invalidated (probably because a loop edge has been removed).
 * `sta_->ensureLevelized()` is called, internally `levelize()` sets level to -1 on a vertex without dequeuing it. Later on `Search::levelChangedBefore` attempts to remove the vertex from the queues but `BfsIterator::remove()` crashes when the vertex level is -1 (it uses a negative index into a vector).

### DIFF
--- a/search/Levelize.cc
+++ b/search/Levelize.cc
@@ -135,7 +135,7 @@ Levelize::levelize()
 
   findRoots();
   findBackEdges();
-  VertexSeq topo_sorted = findToplologicalOrder();
+  VertexSeq topo_sorted = findTopologicalOrder();
   assignLevels(topo_sorted);
   ensureLatchLevels();
 
@@ -345,7 +345,7 @@ Levelize::findUnvisitedVertices()
 ////////////////////////////////////////////////////////////////
 
 VertexSeq
-Levelize::findToplologicalOrder()
+Levelize::findTopologicalOrder()
 {
   Stats stats(debug_, report_);
   std::map<Vertex*, int> in_degree;
--- a/search/Levelize.cc
+++ b/search/Levelize.cc
@@ -130,7 +130,7 @@ Levelize::levelize()
     vertex->setVisited(false);
     vertex->setOnPath(false);
     // assignLevels init
-    vertex->setLevel(-1);
+    setLevel(vertex, -1);
   }
 
   findRoots();
--- a/search/Levelize.hh
+++ b/search/Levelize.hh
@@ -74,7 +74,7 @@ protected:
   void levelize();
   void findRoots();
   VertexSeq sortedRootsWithFanout();
-  VertexSeq findToplologicalOrder();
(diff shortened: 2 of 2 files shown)


akashlevy:
Do you have a standalone example of the crash?

jjcherry56:
The problem with this "fix" is that it glosses over the real issue: a vertex that is not assigned a level after the level is inited to -1 at the start of the levelization.  This:

  // Set level of stranded vertices (constants) to zero.
  VertexIterator vertex_iter2(graph_);
  while (vertex_iter2.hasNext()) {
    Vertex *vertex = vertex_iter2.next();
    if (vertex->level() == -1)
      setLevel(vertex, 0);
 
should set any level that wasn't set to 0 and dequeue it from the arrival_iter. So somethiing else is broken, which means just pasting a band-aid on the problem without a test case doesn't work for me.

povik:
> should set any level that wasn't set to 0 and dequeue it from the arrival_iter.

@jjcherry56 yes, that's the point of crash, look at the code of `BfsIterator::remove()` and consider the vertex having level -1 at that point

As I wrote originally:

> Later on Search::levelChangedBefore attempts to remove the vertex from the queues but BfsIterator::remove() crashes when the vertex level is -1 (it uses a negative index into a vector).

povik:
Let me try once more: if we look at `BfsIterator::remove`
```
void
BfsIterator::remove(Vertex *vertex)
{
  // If the iterator has not been inited the queue will be empty.
  Level level = vertex->level();
  if (vertex->bfsInQueue(bfs_index_)
      && static_cast<Level>(queue_.size()) > level) {
    for (auto &v : queue_[level]) {
      if (v == vertex) {
	v = nullptr;
	vertex->setBfsInQueue(bfs_index_, false);
        break;
      }
    }
  }
}
```
We see that it uses the vertex level to index into `queue_`, which is a zero-based vector. If we set the level to -1 first and dequeue second, we will be accessing `queue[-1]`. This is the real issue my fix resolves. I don't know how to make this any clearer.

jjcherry56:
I understand.
This is why I vastly prefer issues with test cases over PRs with english descriptions.

I built a test case to reproduce the problem and fixed it differently.
This works, but it ends up uses the callback on every vertex in the initialization loop, 
which makes the subsequent calls superfluous. 

see
b32eed9a Fix levelize crash when vertices enqueued in arrival/require

povik:
Thanks! Yes, I went for the minimal change which fixes the problem.

PR #244, opened 2025-05-14
CMake output lib/exe fixes

### LINKED ISSUE #167: CMAKE_ARCHIVE_OUTPUT_DIRECTORY prevents building in multiple directories
This line in CMakeLists.txt:

https://github.com/parallaxsw/OpenSTA/blob/87418f23b1f5c02b0e3334023aedabfefd9b8025/CMakeLists.txt#L480

It makes it so it's cumbersome to build multiple configurations of OpenSTA in the same worktree using different cmake build directories. This is useful for a couple purposes:

* Different cmake configurations (Debug/Release)
* Emulating different architectures (macOS x86-64/aarch64)

The line appears to be a vestige from an older build system, so I wonder if it can be removed. When I tried removing it, all that has had happened is that libOpenSTA.a is in the build directory instead of `app`.

akashlevy:
Fixes #167


CMakeLists.txt
```diff
@@ -553,13 +550,18 @@ target_compile_options(OpenSTA
 set_target_properties(OpenSTA PROPERTIES CXX_EXTENSIONS OFF)
 target_compile_features(OpenSTA PUBLIC cxx_std_17)
 
+# Copy library to app directory.
+add_custom_command(
```
maliberty, line 554: What not use [install](https://cmake.org/cmake/help/latest/command/install.html)?

akashlevy: Because @jjcherry56 specifically requested that the previous behavior stay intact in #167.

Install requires the `make install` target to be used, while this performs the copy with `make` only.

Unless there is some functionality of `install` that I am missing here...

maliberty: That makes sense.

CMakeLists.txt
```diff
@@ -570,6 +572,14 @@ target_link_libraries(sta
   OpenSTA
   )
 
+# Copy executable to app directory.
+add_custom_command(
+  TARGET sta
+  POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E make_directory ${STA_HOME}/app
```
jjcherry56, line 579: that doesn't make any sense because the app directory is in the repo

CMakeLists.txt
```diff
@@ -553,13 +550,18 @@ target_compile_options(OpenSTA
 set_target_properties(OpenSTA PROPERTIES CXX_EXTENSIONS OFF)
 target_compile_features(OpenSTA PUBLIC cxx_std_17)
 
+# Copy library to app directory.
+add_custom_command(
+  TARGET OpenSTA
+  POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E make_directory ${STA_HOME}/app
```
jjcherry56, line 557: that doesn't make any sense because the app directory is in the repo

akashlevy: Ah yeah, this can be removed... sorry about that

jjcherry56:
I'm ok with moving the executable if you do the rest of the work or updating README.md, test/regression.tcl, and documenting the change in doc/ChangeLog.

akashlevy:
What's your preference? I'm happy to move it if you think that's nicer long-term

akashlevy:
Also that's a breaking change, so should ideally be in a minor release if following semver

jjcherry56:
ok, change the version in the cmake too. it will have fallout but the executable really does belong in the build directory

akashlevy:
Cool, I'll bump minor version to 2.7.0 then

akashlevy:
See now, lmk if any more changes are needed

akashlevy:
Sorry, actually done now...

akashlevy:
See now, lmk if any more changes are needed

doc/ApiChanges.txt
```diff
@@ -24,7 +24,13 @@
 
 This file summarizes STA API changes for each release.
 
-Release 2.6.2 2024/03/30
+Release 2.7.0 2025/05/14
```
jjcherry56, line 27: I don't think ApiChanges needs to change.

doc/ApiChanges.txt
```diff
@@ -24,7 +24,13 @@
...
+------------------------
+
+The OpenSTA library `libOpenSTA` and `sta` executable are now built into the
+build directory instead of `app/`.
```
jjcherry56, line 31: built in the

doc/ChangeLog.txt
```diff
@@ -3,7 +3,13 @@ OpenSTA Timing Analyzer Release Notes
 
 This file summarizes user visible changes for each release.
 
-Realase 2.6.1 2025/03/30
+Release 2.7.0 2025/05/14
+-------------------------
+
+The OpenSTA library `libOpenSTA` and `sta` executable are now built into the
```
jjcherry56, line 9: built in the

PR #268, opened 2025-07-10
tcl: readline is only initialized in interactive mode

oharboe:
consistent with OpenROAD

### DIFF
--- a/app/Main.cc
+++ b/app/Main.cc
@@ -122,12 +122,16 @@ staTclAppInit(int argc,
   if (Tcl_Init(interp) == TCL_ERROR)
     return TCL_ERROR;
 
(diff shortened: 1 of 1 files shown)


jjcherry56:
makes sense but I am curious what the functional difference was that you experienced

oharboe:
> makes sense but I am curious what the functional difference was that you experienced

I ran into some tcl errors when building/running with bazel and I discovered that this was not consistent across OpenROAD and OpenSTA, which I found [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and thought it would be good to make consistent as a point of order.

oharboe:
@jjcherry56 Merge or close?

PR #270, opened 2025-07-18
Allow messages.txt to be generated also when OpenSTA makefile is not the top module.

titan73:
Like in OpenROAD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -600,6 +600,6 @@ add_custom_target(sta_tags etags -o TAGS
 add_custom_command(
   TARGET OpenSTA
   POST_BUILD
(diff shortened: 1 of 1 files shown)


titan73:
@jjcherry56 Yes, it currently fails in OpenROAD. This PR is intended to fix this very issue. OpenROAD haven't yet been updated to use lastest OpenSTA so it couldn't take benefit of the fix and thus fails. The OpenROAD's OpenSTA is from 3 weeks ago while the PR was merged 5 days ago.
To fix this, the commit should be re-reverted and a request to update OpenROAD to latest OpenSTA should be done.
I can make this request once the commit is re-reverted.

Issue #277, opened 2025-07-29
Liberty issue: Backslash-newline should be allowed in complex attributes


akashlevy:
OpenSTA warnings:
```
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 56, capacitive_load_units scale is not a float.
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 88, voltage_map voltage is not a float.
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 89, voltage_map voltage is not a float.
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 90, voltage_map voltage is not a float.
```

Line 56:
```
    capacitive_load_unit("1.000000", \
        "ff");
```

Line 88:
```
    voltage_map("VDD", \
        "0.65");
```

Line 90:
```
    voltage_map("VSS", \
        "0");
```

akashlevy:
(As seen in a commercial library with a PrimeTime extracted model)

akashlevy:
Another user reported a more serious occurrence:
```
ERROR: liberty.lib line 49, syntax error
```

Line 49:
```
receiver_capacitance_rise_threshold_pct\
       ( "0, 30.0, 50.0, 60.0, 70.0, 80.0, 90.0, 95.0, 100.00" );
   receiver_capacitance_fall_threshold_pct\
       ( "100, 70.0, 50.0, 40.0, 30.0, 20.0, 10.0, 5.0, 0.0" );
```

This should be supported as well

akashlevy:
Upon further inspection, these are two separate issues. The issue with the `voltage_map` is that the float is encapsulated with quotes. The backslash-newline is a separate issue (very easy to fix). I'll close this issue and split it into two issues

PR #280, opened 2025-07-31
Allow Liberty floats as strings for `voltage_map` and `capacitive_load_unit`

akashlevy:
As mentioned in #277, I found a commercial Liberty library, which had the following OpenSTA warnings:
```
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 56, capacitive_load_units scale is not a float.
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 88, voltage_map voltage is not a float.
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 89, voltage_map voltage is not a float.
[2025-07-27 07:08:51] [STA-INFO] Warning: liberty.lib.gz line 90, voltage_map voltage is not a float.
```

Line 56:
```
    capacitive_load_unit("1.000000", \
        "ff");
```

Line 88:
```
    voltage_map("VDD", \
        "0.65");
```

Line 90:
```
    voltage_map("VSS", \
        "0");
```

Since the quoted floats are recognized as strings, they don't get processed properly. For example, the capacitive load unit stays as the default 1 pF instead of 1 fF.

This PR fixes this by checking if the string value is a float and converting it if so. I added a simple test case as well.


liberty/LibertyReader.cc
```diff
@@ -873,8 +875,19 @@ LibertyReader::visitCapacitiveLoadUnit(LibertyAttr *attr)
       LibertyAttrValueIterator value_iter(attr->values());
       if (value_iter.hasNext()) {
 	LibertyAttrValue *value = value_iter.next();
-	if (value->isFloat()) {
-	  float scale = value->floatValue();
+	float scale = std::nanf("");
```
jjcherry56: I'd rather see you use a bool exists than nan here. std::optional is another option but I find it too clumsy.

akashlevy: Sure, will do that. I thought nan was clever because it is actually not a number

PR #281, opened 2025-08-04
Remove stray printf in BfsIterator::remove

maliberty:
This appears to be leftover debug code:
  printf("bfs remove %s\n", vertex->to_string(this).c_str());


akashlevy:
Seems resolved in https://github.com/parallaxsw/OpenSTA/commit/12f03395ec80d3593f4796b2a3cf5480e75735bd

PR #283, opened 2025-08-12
Change DmpAlg to use preallocated arrays

calewis:
DmpAlg isn't called with nr_order_ > 3 so this change saves 9 news and 9 frees for each construction and destruction of the class.


dcalc/DmpCeff.cc
```diff
@@ -288,27 +293,14 @@ DmpAlg::DmpAlg(int nr_order,
   c1_(0.0),
   nr_order_(nr_order)
 {
-  x_ = new double[nr_order_];
-  fvec_ = new double[nr_order_];
-  scale_ = new double[nr_order_];
-  p_ = new double[nr_order_];
-  fjac_ = new double*[nr_order_];
+  // How do you want to check this?
+  assert(nr_order_ <= max_nr_order_);
```
calewis: Not sure how OpenSTA usually checks things like this.  Let me know and I'll update it to the preferred method.

jjcherry56: Never with assert (tag search or grep for assert and you will see that yourself). You can answer your own question by reading the other code in the same file. As an exercise I want you to explain why the check is not necessary.

calewis: Thanks for the feedback.  I've changed the function to throw a `DmpError` if the order is greater than `max_nr_order_`.   I've taken another look at the code, and I'm still having trouble seeing what makes the check unnecessary.

jjcherry56:
Look at the DmpAlg constructor calls

jjcherry56:
I would agree with the bounds check in general, but in this case the algorithm is very specific so it is pointless. Nobody is going to add another variable to the algorithm without deep enough knowledge to understand the matrix bounds.

PR #289, opened 2025-08-19
Fix an undefined behaviour and a memory leak

Faholan:
Fix an UB (use of memcopy instead of memmove) and a memory leak

### DIFF
--- a/search/Sta.cc
+++ b/search/Sta.cc
@@ -2069,9 +2069,9 @@ Sta::checkExceptionToPins(ExceptionTo *to,
 			  int line) const
 {
   if (to) {
-    PinSet::Iterator (to->pins());
-    while (.hasNext()) {
-      const Pin *pin = .next();
+    PinSet::Iterator pin_iter(to->pins());
+    while (pin_iter.hasNext()) {
+      const Pin *pin = pin_iter.next();
       if (!sdc_->isExceptionEndpoint(pin)) {
 	if (line)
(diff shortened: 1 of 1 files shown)


search/Sta.cc
```diff
@@ -2069,9 +2069,9 @@ Sta::checkExceptionToPins(ExceptionTo *to,
 			  int line) const
 {
   if (to) {
-    PinSet::Iterator pin_iter(to->pins());
-    while (pin_iter.hasNext()) {
-      const Pin *pin = pin_iter.next();
+    PinSet::Iterator (to->pins());
+    while (.hasNext()) {
```
jjcherry56: This change is gratuitous and uses some weird syntax I am not familiar with or really approve of.

search/Sta.cc
```diff
@@ -2953,7 +2953,7 @@ Sta::netSlack(const Net *net,
 {
   ensureGraph();
   Slack slack = MinMax::min()->initValue();
-  NetConnectedPinIterator *pin_iter = network_->connectedPinIterator(net);
+  std::unique_ptr<NetConnectedPinIterator> pin_iter(network_->connectedPinIterator(net));
```
jjcherry56: You may notices the absence of unique_ptr in OpenSTA. That is not an accident.
Please just add the the delete like every other call to a network iterator.

search/Sta.cc
Faholan: Oh that must've been my linter that added that syntax, it wasn't intended I will revert it

PR #293, opened 2025-09-01
Fix leaked iterator in Sta::netSlack

maliberty:


jjcherry56:
see 849832a9 Fix an undefined behaviour and a memory leak (#289)

PR #298, opened 2025-09-15
Add PropertyRegistry for Clock.

mikesinouye:


search/Property.cc
```diff
@@ -1376,6 +1382,14 @@ Properties::defineProperty(std::string &property,
...
 }
 
+void
+Properties::defineProperty(std::string &property,
+                           PropertyRegistry<const Clock *>::PropertyHandler handler)
+{
+  registry_clock_.defineProperty(property, handler);
+}
+
+
```
jjcherry56: Please remove the extra blank line and it is good to go

mikesinouye: Done.

PR #300, opened 2025-09-23
Mark choice of delay implementation with IWYU export pragma.

hzeller:
The Delay header is meant to provide the Delay implementation to whoever is including it; it chooses the right implementation via an include which this PR marks as providing a symbol that is to be exported.

Without that annotation, tools such as `clang-tidy` or the `clangd` language server (as well as many other tools) will complain about headers not directly providing a symbol if users just include Delay.hh; With this annotation, they know.

Documentation about these IWYU pragmas:
https://clangd.llvm.org/design/include-cleaner#iwyu-pragmas https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-begin_exportsend_exports

### DIFF
--- a/include/sta/Delay.hh
+++ b/include/sta/Delay.hh
@@ -26,6 +26,7 @@
 
 #include "StaConfig.hh"
 
+// IWYU pragma: begin_exports
 #if (SSTA == 1)
   // Delays are Normal PDFs.
   #include "DelayNormal1.hh"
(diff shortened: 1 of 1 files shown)


hzeller:
Looks good ?

jjcherry56:
It is innocuous enough, but I think you are over selling it. "Include Cleaner is available in “preview” mode with an incomplete set of capabilities". That hardly sounds like a standard used by "many other tools".

PR #301, opened 2025-09-23
Fix crash in PropActivityVisitor::visit when accessing null LibertyPort

maliberty:


maliberty:
The sequential if was taking care of the case that `test_cell->findLibertyPort(port->name())` returns null.  If that can't happen then it is redundant.

PR #304, opened 2025-09-28
Support more useful properties

akashlevy:
Adding several useful properties onto STA objects for querying/filtering.

### DIFF
--- a/include/sta/Network.hh
+++ b/include/sta/Network.hh
@@ -313,6 +313,9 @@ public:
 		const Instance *hier_inst) const;
   bool isDriver(const Pin *pin) const;
   bool isLoad(const Pin *pin) const;
+  bool isClock(const Pin *pin) const;
+  bool isRiseEdgeTriggered(const Pin *pin) const;
+  bool isFallEdgeTriggered(const Pin *pin) const;
   // Has register/latch rise/fall edges from pin.
   bool isRegClkPin(const Pin *pin) const;
   // Pin clocks a timing check.
--- a/liberty/Liberty.cc
+++ b/liberty/Liberty.cc
@@ -1077,6 +1077,18 @@ LibertyCell::setDontUse(bool dont_use)
   dont_use_ = dont_use;
 }
 
+bool
+LibertyCell::isPhysicalOnly() const
+{
+  // This is an approximation via typical features of a physical-only cell.
+  // The correct way to do this would be to load this info from a LEF file.
+
+  // Assumption: Diodes are not counted for the purposes of this property.
+
+  LibertyCellPortIterator port_it(this); // only iterates on signal ports
+  return !port_it.hasNext(); // zero ports
+}
+
 void
 LibertyCell::setIsMacro(bool is_macro)
 {
@@ -1165,6 +1177,12 @@ void
 LibertyCell::setClockGateType(ClockGateType type)
 {
   clock_gate_type_ = type;
+}
+
+const char *
+LibertyCell::getDesignType() const
+{
+  return isMacro() ? "macro": "cell";
 }
 
 bool
@@ -1478,6 +1496,13 @@ size_t
 LibertyCell::timingArcSetCount() const
 {
   return timing_arc_sets_.size();
+}
+
+bool
+LibertyCell::hasTimingArcs() const
+{
+  return !timing_arc_set_from_map_.empty()
+    || !timing_arc_set_to_map_.empty();
 }
 
 bool
--- a/liberty/Liberty.i
+++ b/liberty/Liberty.i
@@ -265,9 +265,18 @@ default_operating_conditions()
 
 %extend LibertyCell {
 const char *name() { return self->name(); }
+bool has_timing_model() { return self->hasTimingArcs(); }
 bool is_leaf() { return self->isLeaf(); }
 bool is_buffer() { return self->isBuffer(); }
 bool is_inverter() { return self->isInverter(); }
+bool is_clock_gate() { return self->isClockGate(); }
+bool is_integrated_clock_gating_cell() { return self->isClockGate(); }
+bool is_memory() { return self->isMemory(); }
+bool is_physical_only() { return self->isPhysicalOnly(); }
+bool is_sequential() { return self->hasSequentials(); }
+bool dont_use() { return self->dontUse(); }
+void set_dont_use() { return self->setDontUse(true); }
+void unset_dont_use() { return self->setDontUse(false); }
 LibertyLibrary *liberty_library() { return self->libertyLibrary(); }
 Cell *cell() { return reinterpret_cast<Cell*>(self); }
 LibertyPort *
--- a/search/Property.cc
+++ b/search/Property.cc
@@ -40,6 +40,8 @@
 #include "Path.hh"
 #include "power/Power.hh"
 #include "Sta.hh"
+#include "Search.hh"
+#include "PathGroup.hh"
 
 namespace sta {
 
@@ -732,11 +734,19 @@ Properties::getProperty(const LibertyCell *cell,
     return PropertyValue(cell->filename());
   else if (property == "library")
     return PropertyValue(cell->libertyLibrary());
+  else if (property == "is_sequential")
+    return PropertyValue(cell->hasSequentials());
+  else if (property == "has_timing_model")
+    return PropertyValue(cell->hasTimingArcs());
   else if (property == "is_buffer")
     return PropertyValue(cell->isBuffer());
-  else if (property =="is_inverter")
+  else if (property == "is_clock_gate" || property == "is_integrated_clock_gating_cell")
+    return PropertyValue(cell->isClockGate());
+  else if (property == "is_inverter")
     return PropertyValue(cell->isInverter());
-  else if (property == "is_memory")
+  else if (property == "is_physical_only")
+    return PropertyValue(cell->isPhysicalOnly());
+  else if (property == "is_memory" || property == "is_memory_cell")
     return PropertyValue(cell->isMemory());
   else if (property == "dont_use")
     return PropertyValue(cell->dontUse());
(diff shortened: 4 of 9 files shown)


liberty/Liberty.cc
```diff
@@ -1077,6 +1077,18 @@ LibertyCell::setDontUse(bool dont_use)
...
+LibertyCell::isPhysicalOnly() const
+{
+  // This is an approximation via typical features of a physical-only cell.
+  // The correct way to do this would be to load this info from a LEF file.
+
+  // Assumption: Diodes are not counted for the purposes of this property.
+
+  LibertyCellPortIterator port_it(this); // only iterates on signal ports
+  return !port_it.hasNext(); // zero ports
+}
```
QuantamHD: Can you make sure this is overridable, because OpenROAD prefers control of this property.

liberty/Liberty.i
```diff
@@ -265,9 +265,18 @@ default_operating_conditions()
...
 bool is_leaf() { return self->isLeaf(); }
 bool is_buffer() { return self->isBuffer(); }
 bool is_inverter() { return self->isInverter(); }
+bool is_clock_gate() { return self->isClockGate(); }
+bool is_integrated_clock_gating_cell() { return self->isClockGate(); }
+bool is_memory() { return self->isMemory(); }
+bool is_physical_only() { return self->isPhysicalOnly(); }
+bool is_sequential() { return self->hasSequentials(); }
+bool dont_use() { return self->dontUse(); }
+void set_dont_use() { return self->setDontUse(true); }
```
gadfort: Why return in `void` functions?

liberty/Liberty.cc
akashlevy: I think it should be overridable as is. I haven't tested with OpenROAD, but is there anything you think needs to be changed?

liberty/Liberty.i
akashlevy: whoops

liberty/Liberty.cc
QuantamHD: Or perhaps implement it as a custom property in your own fork?

akashlevy: Makes sense, I can leave this one downstream

QuantamHD: Thanks @akashlevy. ❤️

include/sta/Network.hh
```diff
@@ -313,6 +313,9 @@ public:
 		const Instance *hier_inst) const;
   bool isDriver(const Pin *pin) const;
   bool isLoad(const Pin *pin) const;
+  bool isClock(const Pin *pin) const;
+  bool isRiseEdgeTriggered(const Pin *pin) const;
+  bool isFallEdgeTriggered(const Pin *pin) const;
```
jjcherry56: I don't think these belong in the network api.
They are not very reliable given the various possible sequential clocked_on expression and certainly don't work with statetables (good lock).

liberty/Liberty.cc
```diff
@@ -1167,6 +1167,12 @@ LibertyCell::setClockGateType(ClockGateType type)
   clock_gate_type_ = type;
 }
 
+const char *
+LibertyCell::getDesignType() const
+{
```
jjcherry56: This is not really complete (pads? clock gates? inverters?), and I am not convinced cell types are that easy to pin down. I would say if you want this it should be on your own end.

liberty/Liberty.i
```diff
@@ -265,9 +265,17 @@ default_operating_conditions()
...
 bool is_leaf() { return self->isLeaf(); }
 bool is_buffer() { return self->isBuffer(); }
 bool is_inverter() { return self->isInverter(); }
+bool is_clock_gate() { return self->isClockGate(); }
+bool is_integrated_clock_gating_cell() { return self->isClockGate(); }
+bool is_memory() { return self->isMemory(); }
+bool is_sequential() { return self->hasSequentials(); }
+bool dont_use() { return self->dontUse(); }
+void set_dont_use() { self->setDontUse(true); }
+void unset_dont_use() { self->setDontUse(false); }
```
jjcherry56: why change the api with set/unset instead of just using one function?

search/Property.cc
```diff
@@ -767,7 +776,26 @@ Properties::getProperty(const Port *port,
     return PropertyValue(network->direction(port)->name());
   else if (property == "liberty_port")
     return PropertyValue(network->libertyPort(port));
-
+  else if (property == "capacitance"
+           || property == "pin_capacitance") {
+    int fanout = 0;
+    float cap = 0.0, wire_cap = 0.0;
+    for (const Corner *corner : *sta_->corners())
+      sta_->portExtCaps(port, corner, MinMax::max(), cap, wire_cap, fanout);
```
jjcherry56: this returns the cap for the "last" corner, which is not very well defined.u
it also returns the SDC external capacitance set by "set_load -pin port", which is not what I would call the capacitance of a liberty port.

liberty/Liberty.cc
akashlevy: Got it, will remove

search/Property.cc
akashlevy: Should probably evaluate at `cmd_corner_` instead?

And what would you suggest instead of the SDC caps? Our idea was to be able to run this command without a full dcalc run

liberty/Liberty.i
akashlevy: Not sure this is needed, can remove

include/sta/Network.hh
akashlevy: We can keep the edge triggers in our fork given your (valid) concerns about completeness.

So `isClock` should just be a function on the pin directly without going through the Network API?

search/Property.cc
akashlevy: We'll just keep this downstream too I think...

liberty/Liberty.i
akashlevy: These commands are supported by many EDA tools, but we can keep it downstream for simplicity

include/sta/Network.hh
akashlevy: I'm leaving isClock in Network API but removing rise/fall-edge triggered since they are hard to determine properly...

liberty/Liberty.i
```diff
@@ -265,9 +265,15 @@ default_operating_conditions()
...
 const char *name() { return self->name(); }
+bool has_timing_model() { return self->hasTimingArcs(); }
 bool is_leaf() { return self->isLeaf(); }
 bool is_buffer() { return self->isBuffer(); }
 bool is_inverter() { return self->isInverter(); }
+bool is_clock_gate() { return self->isClockGate(); }
+bool is_integrated_clock_gating_cell() { return self->isClockGate(); }
+bool is_memory() { return self->isMemory(); }
+bool is_sequential() { return self->hasSequentials(); }
+bool dont_use() { return self->dontUse(); }
```
jjcherry56: why do you need these AND properties?
I have migrated toward using get_property over swig object syntax.

search/Property.cc
```diff
@@ -1260,14 +1277,26 @@ Properties::getProperty(PathEnd *end,
     PathExpanded expanded(end->path(), sta_);
     return PropertyValue(expanded.startPath()->pin(sta_));
   }
-  else if (property == "startpoint_clock")
-    return PropertyValue(end->path()->clock(sta_));
+  else if (property == "startpoint_clock"){
+    const Clock *clk = end->path()->clock(sta_);
+    if (clk)
+      return PropertyValue(clk->name());
```
jjcherry56, line 1285: why change it to the clock name? you can't get properties of a clock name like a clock object

search/Property.cc
```diff
@@ -1260,14 +1277,26 @@ Properties::getProperty(PathEnd *end,
...
+      return PropertyValue();
+  }
   else if (property == "endpoint")
     return PropertyValue(end->path()->pin(sta_));
-  else if (property == "endpoint_clock")
-    return PropertyValue(end->targetClk(sta_));
+  else if (property == "endpoint_clock") {
+    const Clock *clk = end->targetClk(sta_);
+    if (clk)
+      return PropertyValue(clk->name());
```
jjcherry56, line 1294: ditto

search/Property.cc
```diff
@@ -1260,14 +1277,26 @@ Properties::getProperty(PathEnd *end,
...
+    else
+      return PropertyValue();
+  }
   else if (property == "endpoint_clock_pin")
     return PropertyValue(end->targetClkPath()->pin(sta_));
+  else if (property == "path_group")
+    return PropertyValue(sta_->search()->pathGroup(end)->name());
```
jjcherry56, line 1301: it is plural now: pathGroups

include/sta/Network.hh
```diff
@@ -313,6 +313,7 @@ public:
 		const Instance *hier_inst) const;
   bool isDriver(const Pin *pin) const;
   bool isLoad(const Pin *pin) const;
+  bool isClock(const Pin *pin) const;
```
jjcherry56: The problem with this is that there are already many isClock functions, each of which have different semantics. I would rather the caller explicitly query the liberty port to know exactly
what they are getting. Network functions like this are convenience futions because they are used in multiple places but this only has one caller.

liberty/Liberty.i
akashlevy: Got it. This was mostly to be in line with the other properties which have both. Can remove.

search/Property.cc
akashlevy: Will fix

include/sta/Network.hh
akashlevy: Ok, I'll explicitly look up the liberty port here and leave this out of the Network API

search/Property.cc
akashlevy: The standard behavior of most EDA tools seems to be to return strings not objects when you request properties.

search/Property.cc
akashlevy: See my comment above.

search/Property.cc
akashlevy: I think ideally we ensure that we are always returning strings when getting properties to be consistent with other tool behavior

search/Property.cc
akashlevy: Hmm, I'm not seeing the API for this. Can you double check?

PR #305, opened 2025-09-28
Include StaConfig.hh once

akashlevy:
This was causing issues with a CentOS 7 configuration with an older compiler/cmake.

### DIFF
--- a/util/StaConfig.hh.cmake
+++ b/util/StaConfig.hh.cmake
@@ -1,5 +1,4 @@
-#ifndef STA_CONFIG_HH
-#define STA_CONFIG_HH
+#pragma once
 
 #define STA_VERSION "${STA_VERSION}"
 
@@ -10,5 +9,3 @@
 #define SSTA ${SSTA}
 
 #define TCL_READLINE ${TCL_READLINE}
-
-#endif


util/StaConfig.hh.cmake
```diff
@@ -1,3 +1,6 @@
+#ifndef STA_CONFIG_HH
```
gadfort: Other headers use `#pragma once` might be good to be consistent

akashlevy: The reason I used `#ifndef` instead of `#pragma once` was to maintain better compatibility with other compilers, since this config file gets used in other projects when you include OpenSTA as a library.

maliberty: What compiler doesn't support pragma once?

akashlevy: I had tried `#pragma once` initially and I'm pretty sure the older CMake version I was using on CentOS 7 had some issue with multiple definitions with that.

This isn't a normal header, it's a CMake template. Pretty much all C/C++ compilers will respect pragma once, but I'm not sure that the older CMake preprocessors will.

I'm going to double check again whether `#pragma once` works with our setup. If it works, I'll update it, else I'll provide evidence that this is needed.

akashlevy: No issues with `#pragma once` it seems. Might have been a result of not rebuilding from scratch after making the change.

PR #306, opened 2025-09-29
Thread local tag cache

calewis:
Add a thread local cache in  of findTag to reduce lock contention.
`findTag` spends ~90% of its time contending on the lock. This pr lowers
that significantly, taking findTag from 3.9% of wall time to 0.8%.

There will be a small amount of Tag* leakage until the threads are
destroyed at which point the thread local caches will be automatically
cleaned up.

### DIFF
--- a/include/sta/Search.hh
+++ b/include/sta/Search.hh
@@ -72,6 +72,8 @@ typedef Vector<VertexSlackMap> VertexSlackMapSeq;
 typedef Vector<WorstSlacks> WorstSlacksSeq;
 typedef std::vector<DelayDbl> DelayDblSeq;
 
+class ThreadLocalCacheTagSet;
+
 class Search : public StaState
 {
 public:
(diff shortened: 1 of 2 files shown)


tspyrou:
This cuts a full run from 8 hrs to 7.5 hrs.

jjcherry56:
I sort of get the idea of what this is doing, but I don't completely understand all of the complications with "clear_counts". My concern is that I get stuck with fixing the non-trivial thread related bugs that show up after this goes in the code.

It needs a LOT of work to look like the other code in OpenSTA, which is another prerequisite. I can't even begin to list them there are so many, so I suggest you read read doc/CodingGuidelines.txt and look around and try to match the existing code.

jjcherry56:
I think the idea of using a cache in front of the lock is a good one,
but I do not believe in using thread_local. 
I think it is cleaner to just use the stack (eg, the requirement for tl_cache_map). In this
case the ArrivalVisitor or TagGroupBldr could hold the cache there a
copies for each sta/thread.

PR #308, opened 2025-10-01
Change DispatchQueue::dispatch to use notify_one.

calewis:
This prevents the thundering herd problem and should increase the scalability of the DispatchQueue significantly.

Additionally the code the DispatchQueue was taken from made this improvement five years ago:
https://github.com/embeddedartistry/embedded-resources/commit/79ad8a539d7e264350464e434e03b5d910ae4a29


jjcherry56:
Is that the only update that needs to be imported?

PR #309, opened 2025-10-03
Recognize some basic specify blocks and ignore them

ambd161:
Some commercial tools spew out specify blocks in structural netlists as there is nothing in the Verilog language spec to forbid that. This PR simply adds recognition of `specify` `endspecify` and `specparam` to the verilog parser so that they can be ignored and OpenROAD can work with netlists directly out of virtuoso without postprocessing.

e.g. this will no longer throw an error but will be ignored by the verilog parser
```verilog
specify 
    specparam CDS_LIBNAME  = "test_lib";
    specparam CDS_CELLNAME = "test_cell";
    specparam CDS_VIEWNAME = "schematic";
endspecify
```

### DIFF
--- a/test/regression_vars.tcl
+++ b/test/regression_vars.tcl
@@ -138,7 +138,6 @@ record_example_tests {
 }
 
 record_sta_tests {
-  verilog_lang
   get_filter
   get_is_memory
   get_lib_pins_of_objects


jjcherry56:
Please add a regression in /test using your verilog snippet.
See the updated doc/CodingGuidelines.txt for instructions.

test/regression_vars.tcl
```diff
@@ -138,6 +138,7 @@ record_example_tests {
 }
 
 record_sta_tests {
+  verilog_lang
```
jjcherry56: You might notice the tests area in alphabetical order (modulo the last PR that screwed up and I missed). The name "verilog_lang" is too generic; every regression reads verilog. I suggested verilog_specify, because that is the structure it is testing.

test/verilog_lang.tcl
```diff
@@ -0,0 +1,12 @@
+# try to load verilog language file
+read_liberty ../examples/sky130hd_tt.lib.gz
+read_verilog verilog_lang.v
```
jjcherry56: read_verilog is the only command required to test the parser. The rest is just noise.

test/verilog_lang.v
```diff
@@ -0,0 +1,33 @@
...
+  parameter PARAM2="test";
+
+  specify
+    specparam SPARAM1=2;
+    specparam SPARAM2="test2";
+  endspecify
+
+  defparam _1415_.PARAM2 = 1;
+
+  sky130_fd_sc_hd__dfrtp_1 _1415_ (
```
jjcherry56: No instances need to be in the verilog

test/regression_vars.tcl
ghost: it was logical for me that there should be a single be-all end-all regression test that checks verilog parsing and processing, and that should be the first thing to validate. I didn't call it verilog_specify, since the PR simply ignores the `specify` block in a manner similar to how `parameter` and `defparam` are also ignored. So while I was at it, I also added those constructs as well to the input 

I'll change as you suggested

PR #316, opened 2025-10-17
Add local tag caches to PathVisitors

calewis:
Replaces #306.

Adds a tag cache to `PathVisitors` that is passed into `thruTag` and propagated to `findTag`.  This can significantly reduce lock contention in `visitParallel`.

I set it up to only create tag caches in the `visitParallel` context since there is no lock contention and thus no need for caching when running visits on the main thread/visitor.

### DIFF
--- a/include/sta/Search.hh
+++ b/include/sta/Search.hh
@@ -753,7 +753,7 @@ protected:
 			       const MinMax *min_max,
 			       const PathAnalysisPt *path_ap) = 0;
   SearchPred *pred_;
-  std::unique_ptr<TagSet> tag_cache_ = nullptr;
+  std::unique_ptr<TagSet> tag_cache_;
 };
 
 // Visitor called during forward search to record an
--- a/search/Search.cc
+++ b/search/Search.cc
@@ -2037,14 +2037,16 @@ Search::inputDelayTag(const Pin *pin,
 
 PathVisitor::PathVisitor(const StaState *sta) :
   StaState(sta),
-  pred_(sta->search()->evalPred())
+  pred_(sta->search()->evalPred()),
+  tag_cache_(nullptr)
 {
 }
 
 PathVisitor::PathVisitor(SearchPred *pred,
 			 const StaState *sta) :
   StaState(sta),
-  pred_(pred)
+  pred_(pred),
+  tag_cache_(nullptr)
 {
 }
 
@@ -2470,12 +2472,10 @@ Search::thruTag(Tag *from_tag,
   const RiseFall *from_rf = from_tag->transition();
   const ClkInfo *from_clk_info = from_tag->clkInfo();
   bool to_is_reg_clk = to_vertex->isRegClk();
-  Tag *to_tag = mutateTag(from_tag, from_pin, from_rf, false, from_clk_info, to_pin, to_rf, false, to_is_reg_clk, false,
+  Tag *to_tag = mutateTag(from_tag, from_pin, from_rf, false, from_clk_info,
+                          to_pin, to_rf, false, to_is_reg_clk, false,
                           // input delay is not propagated.
-                          from_clk_info,
-                          nullptr,
-                          min_max,
-                          path_ap,
+                          from_clk_info, nullptr, min_max, path_ap,
                           tag_cache);
   return to_tag;
 }
@@ -2725,7 +2725,9 @@ Search::mutateTag(Tag *from_tag,
     sdc_->exceptionThruStates(from_pin, to_pin, to_rf, min_max, new_states);
 
   if (new_states)
-    return findTag(to_rf, path_ap, to_clk_info, to_is_clk, from_tag->inputDelay(), to_is_segment_start, new_states, true, tag_cache);
+    return findTag(to_rf, path_ap, to_clk_info, to_is_clk,
+                   from_tag->inputDelay(), to_is_segment_start, new_states,
+                   true, tag_cache);
   else {
     // No state change.
     if (to_clk_info == from_clk_info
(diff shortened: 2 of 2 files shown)


search/Search.cc
```diff
@@ -361,6 +361,10 @@ Search::deleteTags()
   tag_set_->deleteContentsClear();
   tag_free_indices_.clear();
 
+  for (auto cache : tag_caches_) {
```
jjcherry56: this loop is pointless because tags are never going to be deleted during a bfs search and tag_caches_ is alway empty.

search/Search.cc
```diff
@@ -2046,6 +2057,23 @@ PathVisitor::PathVisitor(SearchPred *pred,
 {
 }
 
+void
+PathVisitor::registerTagCache()
```
jjcherry56: So since there are no uses of tag_caches_ in Search the register/deregister functions are pointless.

search/Bfs.cc
```diff
@@ -171,13 +172,17 @@ BfsIterator::visitParallel(Level to_level,
...
   int visit_count = 0;
+  bool is_path_visitor = dynamic_cast<PathVisitor *>(visitor) != nullptr;
   if (!empty()) {
     if (thread_count == 1)
       visit_count = visit(to_level, visitor);
     else {
       std::vector<VertexVisitor*> visitors;
-      for (int k = 0; k < thread_count_; k++)
-	visitors.push_back(visitor->copy());
+      for (int k = 0; k < thread_count_; k++) {
```
jjcherry56: The caches should not require any changes to Bfs.cc.
That is what class member functions are for.
This is a pretty flagrant violation of the api. Any time you start doing dynamic_cast/if object something is probably wrong.

search/Search.cc
```diff
@@ -2046,6 +2057,23 @@ PathVisitor::PathVisitor(SearchPred *pred,
...
+{
+  if (!search_)
```
jjcherry56: how could there be no search_? This is pointless (but then the whole function is also).

search/Search.cc
```diff
@@ -2046,6 +2057,23 @@ PathVisitor::PathVisitor(SearchPred *pred,
...
+    return;
+  if (!tag_cache_)
+    tag_cache_ = std::make_unique<TagSet>(128, TagSet::hasher(this), TagSet::key_equal(this));
+
+  search_->registerTagCache(tag_cache_.get());
+}
+
+PathVisitor::~PathVisitor()
+{
+  if (tag_cache_ && search_)
```
jjcherry56: Again, you can't have an STA without a search_ so the && search_ is pointless.

include/sta/Search.hh
```diff
@@ -749,6 +753,7 @@ protected:
 			       const MinMax *min_max,
 			       const PathAnalysisPt *path_ap) = 0;
   SearchPred *pred_;
+  std::unique_ptr<TagSet> tag_cache_ = nullptr;
```
jjcherry56: init in constructor like all the other members

search/Search.cc
```diff
@@ -2450,21 +2458,25 @@ Search::clkInfoWithCrprClkPath(const ClkInfo *from_clk_info,
...
   Vertex *to_vertex = edge->to(graph_);
   const Pin *to_pin = to_vertex->pin();
   const RiseFall *from_rf = from_tag->transition();
   const ClkInfo *from_clk_info = from_tag->clkInfo();
   bool to_is_reg_clk = to_vertex->isRegClk();
-  Tag *to_tag = mutateTag(from_tag, from_pin, from_rf, false, from_clk_info,
-			  to_pin, to_rf, false, to_is_reg_clk, false,
-			  // input delay is not propagated.
-			  from_clk_info, nullptr, min_max, path_ap);
+  Tag *to_tag = mutateTag(from_tag, from_pin, from_rf, false, from_clk_info, to_pin, to_rf, false, to_is_reg_clk, false,
```
jjcherry56: restore formatting so there isn't an excessively long line and a bunch of 1 arg lines

search/Search.cc
```diff
@@ -2712,9 +2725,7 @@ Search::mutateTag(Tag *from_tag,
     sdc_->exceptionThruStates(from_pin, to_pin, to_rf, min_max, new_states);
 
   if (new_states)
-    return findTag(to_rf, path_ap, to_clk_info, to_is_clk,
-		   from_tag->inputDelay(), to_is_segment_start,
-		   new_states, true);
+    return findTag(to_rf, path_ap, to_clk_info, to_is_clk, from_tag->inputDelay(), to_is_segment_start, new_states, true, tag_cache);
```
jjcherry56: restore formatting so there isn't an excessively long line

search/Search.cc
```diff
@@ -2724,9 +2735,7 @@ Search::mutateTag(Tag *from_tag,
 	&& from_tag->inputDelay() == to_input_delay)
       return from_tag;
     else
-      return findTag(to_rf, path_ap, to_clk_info, to_is_clk,
-		     to_input_delay, to_is_segment_start,
-		     from_states, false);
+      return findTag(to_rf, path_ap, to_clk_info, to_is_clk, to_input_delay, to_is_segment_start, from_states, false, tag_cache);
```
jjcherry56: restore formatting so there isn't an excessively long line

calewis: Would you like to add a column limit to the .clang-format file so that the formatter will fix that?  This very long line came from `ColumnLimit: 0`, which means no width limit.

include/sta/Search.hh
```diff
@@ -753,7 +753,7 @@ protected:
 			       const MinMax *min_max,
 			       const PathAnalysisPt *path_ap) = 0;
   SearchPred *pred_;
-  std::unique_ptr<TagSet> tag_cache_ = nullptr;
+  std::unique_ptr<TagSet> tag_cache_;
 };
```
jjcherry56: I understand the desire to use unique_ptr, but they are not used anywhere else in OpenSTA and I don't see much point in being inconsistent. shared_ptr is a more useful beast.

calewis:
@jjcherry56 Just need to know what else you'd like to see.

PR #317, opened 2025-10-21
Update the maximum line width in .clang-format

calewis:
Update .clang-format based on comments in #316

### DIFF
--- a/search/Bfs.cc
+++ b/search/Bfs.cc
@@ -1,25 +1,25 @@
 // OpenSTA, Static Timing Analyzer
 // Copyright (c) 2025, Parallax Software, Inc.
-// 
+//
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
 // the Free Software Foundation, either version 3 of the License, or
 // (at your option) any later version.
-// 
+//
 // This program is distributed in the hope that it will be useful,
 // but WITHOUT ANY WARRANTY; without even the implied warranty of
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 // GNU General Public License for more details.
-// 
+//
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
-// 
+//
 // The origin of this software must not be misrepresented; you must not
 // claim that you wrote the original software.
-// 
+//
 // Altered source versions must be plainly marked as such, and must not be
 // misrepresented as being the original software.
-// 
+//
 // This notice may not be removed or altered from any source distribution.
 
 #include "Bfs.hh"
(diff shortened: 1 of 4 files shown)


jjcherry56:
If you run clang-format and look at the diffs you can see all the places it fails. 
I haven't checked in a few years because I gave up on it. Half of the options just
didn't work at all.

calewis:
@jjcherry56 I made some changes that fix at least some of your issues and I tried it on Search.{hh,cc} and Bfs.cc.  Is there still stuff you don't like?

jjcherry56:
I don't have time to deal with this right now

search/Bfs.cc
```diff
@@ -363,42 +359,45 @@ BfsFwdIterator::incrLevel(Level &level) const
...
 
 ////////////////////////////////////////////////////////////////
 
 BfsBkwdIterator::BfsBkwdIterator(BfsIndex bfs_index,
-				 SearchPred *search_pred,
-				 StaState *sta) :
-  BfsIterator(bfs_index, level_max, 0, search_pred, sta)
+                                 SearchPred *search_pred,
+                                 StaState *sta) :
+  BfsIterator(bfs_index,
```
jjcherry56, line 396: I could live with this but I don't understand why these args are on separate lines when other function calls are broken by line width.

search/Bfs.cc
```diff
@@ -37,10 +37,10 @@
 namespace sta {
 
 BfsIterator::BfsIterator(BfsIndex bfs_index,
-			 Level level_min,
-			 Level level_max,
-			 SearchPred *search_pred,
-			 StaState *sta) :
+                         Level level_min,
```
jjcherry56, line 40: I agree with untabification, but the problem is I have some branches with very extensive changes that will make merge/rebase next to impossible unless I can come up with a strategy to do them both at the same time so there are no conflicts.

search/Bfs.cc
```diff
@@ -68,9 +68,7 @@ BfsIterator::ensureSize()
   }
 }
 
-BfsIterator::~BfsIterator()
-{
-}
+BfsIterator::~BfsIterator() {}
```
jjcherry56, line 71: weirdness

search/Bfs.cc
```diff
@@ -417,34 +416,33 @@ BfsBkwdIterator::incrLevel(Level &level) const
...
   if (search_pred->searchTo(vertex)) {
     VertexInEdgeIterator edge_iter(vertex, graph_);
     while (edge_iter.hasNext()) {
       Edge *edge = edge_iter.next();
       Vertex *from_vertex = edge->from(graph_);
-      if (from_vertex->level() >= to_level
-	  && search_pred->searchFrom(from_vertex)
-	  && search_pred->searchThru(edge))
-	enqueue(from_vertex);
+      if (from_vertex->level() >= to_level && search_pred->searchFrom(from_vertex)
```
jjcherry56, line 441: I really like having the boolean clauses broken by lines, but it may not be smart enough to do that.

PR #329, opened 2025-11-10
Fix `isBuffer`

akashlevy:
I discovered a bug when one of our tests failed on `is_buffer` property check on SKY130. It's clearly a buffer but is marked as a non-buffer because of the wells in SKY130. I've included a minimal example in the `tests/` directory.

This PR adds a `PortDirection` enum value for wells (`nwell`, `pwell`, `deepnwell`, `deeppwell`) and ignores them in a couple places where they should be ignored (I just searched for `isPowerGround()` and evaluated whether wells should be ignored at those places).

Let me know if any changes are needed. Thanks!

### DIFF
--- a/include/sta/PortDirection.hh
+++ b/include/sta/PortDirection.hh
@@ -41,7 +41,6 @@ public:
   static PortDirection *internal() { return internal_; }
   static PortDirection *ground() { return ground_; }
   static PortDirection *power() { return power_; }
-  static PortDirection *well() { return well_; }
   static PortDirection *unknown() { return unknown_; }
   static PortDirection *find(const char *dir_name);
   const char *name() const { return name_; }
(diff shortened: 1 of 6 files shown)


akashlevy:
The difference between `isPowerGround` and `isPwrGnd` was confusing me. I didn't notice `isPwrGnd` before, that seems to obviate the need for distinguishing wells.

Probably the best solution is to simply replace most instances of `isPowerGround` (which matches only power and ground) with `isPwrGnd` (which matches any valid pg pin). Shall I amend this to do that?

maliberty:
> (Also `isPwrGnd` could be renamed to is `isPGPin` or something like that for more clarity)

I can't say I find that clearer and we are in the process of integrating with the current name.

parallaxsw:
It could. But in OpenSTA speak it is not a pin; it is a port. And there is
no point in having a Port in the name of a port member function because it
is redundant. The header comment tries to make the pg_pin related functions
clear.

akashlevy:
Yeah, let's minimize changes; I don't care to argue for `well` as a port direction or `isPwrGnd` naming.

Let's just fix the issue at hand, which is `isBuffer`. One-line fix with simple test case demonstrating the issue

PR #330, opened 2025-11-17
Add missing system header includes.

Logikable:


Logikable:
we received an internal bug report from the C++ team as part of a new toolchain release. Not entirely sure what changed

PR #348, opened 2025-12-04
Large bus width bug with read_vcd

stanminlee:
For large bus widths, read_vcd does not work because it converts the binary string token into an integer. This leaves the possibility of an overflow and an incorrect annotation for each pin along the bus. Reference the issue I created for more specifics.

My solution uses the token itself to annotate the pins.


power/VcdParse.hh
```diff
@@ -118,7 +118,7 @@ public:
                               char value) = 0;
   virtual void varAppendBusValue(const std::string &id,
                                  VcdTime time,
-                                 int64_t bus_value) = 0;
+                                 const char* bus_value_str) = 0;
```
jjcherry56: use std::string & instead of char* and do not rename the arg (the type tells you the type so there is no point in repeating it)

stanminlee: Resolved, thanks!

power/VcdParse.cc
```diff
@@ -243,15 +244,15 @@ VcdParse::parseVarValues()
...
-        char *end;
-        int64_t bus_value = strtol(bin.c_str(), &end, 2);
+        string bus_value = token.substr(1);
         string id = getToken();
         if (!reader_->varIdValid(id))
           report_->fileError(807, filename_, stmt_line_,
                              "unknown variable %s", id.c_str());
-        else
+        else {
+          std::reverse(bus_value.begin(), bus_value.end()); // reverse the bus value to match the bit order in the VCD file.
```
jjcherry56: please move the comment to the line before so it isn't so long.

stanminlee: done

PR #359, opened 2026-01-08
Use Map with a deterministic instance power iterator

mikesinouye:
In the instance power summation, the iteration order is non-deterministic as the map uses key Instance*. This results in minor non-determinism of final power values, since the order of the floating point accumulates differ. This change uses the existing InstanceIdLess CMP to remove report_power non-determinism.

https://github.com/parallaxsw/OpenSTA/blob/master/power/Power.cc#L301-L307

### DIFF
--- a/power/Power.hh
+++ b/power/Power.hh
@@ -33,7 +33,6 @@
 #include "PowerClass.hh"
 #include "StaState.hh"
 #include "Bdd.hh"
-#include "Map.hh"
 
 struct DdNode;
 struct DdManager;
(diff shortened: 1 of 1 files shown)


power/Power.hh
```diff
@@ -238,7 +239,7 @@ private:
   PwrSeqActivityMap seq_activity_map_;
   bool activities_valid_;
   Bdd bdd_;
-  std::map<const Instance*, PowerResult> instance_powers_;
+  Map<const Instance*, PowerResult, InstanceIdLess> instance_powers_;
```
jjcherry56: I am phasing out sta::Map so please update it to use std::map

PR #380, opened 2026-02-11
Expose 'TimingType' from `TimingArc`.

rocallahan:
When using the OpenSTA Liberty file parser as a general-purpose Liberty reader, it's useful to be able to extract the original `TimingType` for a timing arc, and trivial to support.


calewis:
@jjcherry56 Rob works with us.  Do you think you'll get a chance to look at this?

Issue #389, opened 2026-02-26
Duplicate timingType() declaration in TimingArcSet causes compilation failure


kazutoiris:
A duplicate declaration of the `timingType()` function exists in the `TimingArcSet` class within the `include/sta/TimingArc.hh` header file, which results in a compilation error during the build process. The identical function signature is declared twice in the public scope of the `TimingArcSet` class with the exact same implementation:

https://github.com/parallaxsw/OpenSTA/blob/f0c9971015015fa9cded7d2c1427bee55718ddfb/include/sta/TimingArc.hh#L155

https://github.com/parallaxsw/OpenSTA/blob/f0c9971015015fa9cded7d2c1427bee55718ddfb/include/sta/TimingArc.hh#L173

This duplicate definition triggers a compiler overload conflict error, as the C++ standard does not allow multiple identical member function declarations with the same signature, return type, and const qualifier in a single class.

```
FAILED: [code=1] CMakeFiles/OpenSTA.dir/app/StaMain.cc.o 
/usr/bin/c++  -I/home/runner/work/artifact/artifact/include -I/usr/include/tcl -I/home/runner/work/artifact/artifact/build/include/sta -I/home/runner/work/artifact/artifact/include/sta -I/home/runner/work/artifact/artifact -I/home/runner/work/artifact/artifact/cudd/include -I/home/runner/work/artifact/artifact/build -isystem /usr/include/eigen3 -O3 -DNDEBUG -std=c++20 -flto=auto -fno-fat-lto-objects -Wall -Wextra -pedantic -Wcast-qual -Wredundant-decls -Wformat-security -Werror=misleading-indentation -Wno-format-zero-length -MD -MT CMakeFiles/OpenSTA.dir/app/StaMain.cc.o -MF CMakeFiles/OpenSTA.dir/app/StaMain.cc.o.d -o CMakeFiles/OpenSTA.dir/app/StaMain.cc.o -c /home/runner/work/artifact/artifact/app/StaMain.cc
In file included from /home/runner/work/artifact/artifact/include/sta/ArcDelayCalc.hh:33,
                 from /home/runner/work/artifact/artifact/include/sta/Sta.hh:43,
                 from /home/runner/work/artifact/artifact/app/StaMain.cc:33:
[...]

PR #390, opened 2026-02-26
Revert duplicate timingType() declaration in TimingArcSet

kazutoiris:
Now the `TimingArcSet` class in `TimingArc.hh` has two identical declarations of the `timingType()` member function with the same signature, return type and implementation, which causes an overload conflict error during the build process.

This reverts commit f0c9971015015fa9cded7d2c1427bee55718ddfb.

## Related Issue
#389 

## Related PR
#380


jjcherry56:
yes, but it is a pain to deal with and I am not interested in working on it really. There is an existing PR to add actions. I never tried to get it working. Go for it. Even giiving you permissions to edit the actions stresses my github foo.

Issue #398, opened 2026-03-05
Hierarchical verilog with unconnected pins


dsengupta0628:
In a hierarchical design containing 2 blocks block1 b1 and block2 b2, but one of the block's (say b2's) 2nd output is unconnected in the parent module, then  get_pins "b2/out2", even if there is a valid net connection.

This testcase can be run by placing both .v and .tcl inside the OpenSTA/test directory

  The Problem:
  - In block2, u3/Z is connected to port out2
  - When b2/out2 is unconnected in the parent, there's no pin for it
  - get_property [get_pins b2/out2] full_name ERRORs
  - Without a pin, the hierarchical boundary is broken
  
Verilog:
```
module top (in, clk1, clk2, out, out2);
   input in, clk1, clk2;
   output out, out2;
   block1 b1 (.in(in), .clk(clk1), .out(b1out), .out2(out2));
   block2 b2 (.in(b1out), .clk(clk2), .out(out));
endmodule // top

module block1 (in, clk, out, out2);
   input in, clk;
   output out, out2;
   BUF_X1 u1 (.A(in), .Z(u1out));
   DFF_X1 r1 (.D(u1out), .CK(clk), .Q(r1q));
   BUF_X1 u2 (.A(r1q), .Z(out));
   BUF_X1 u3 (.A(out), .Z(out2));
endmodule // block1

module block2 (in, clk, out, out2);
   input in, clk;
   output out, out2;
   BUF_X1 u1 (.A(in), .Z(u1out));
   DFF_X1 r1 (.D(u1out), .CK(clk), .Q(r1q));
   BUF_X1 u2 (.A(r1q), .Z(out));
   BUF_X1 u3 (.A(out), .Z(out2));
endmodule // block2
```

TCL script:
```
read_liberty ../examples/nangate45_typ.lib.gz
read_verilog unconnected_dbterm.v
link_design top 
puts "Find b1/out2: [get_property [get_pins b1/out2] full_name]"
puts "Find b2/out2: [get_property [get_pins b2/out2] full_name]"
# Check if net is connected to "b2/u3/Z" 
set iterm [sta::find_pin "b2/u3/Z"]
set net [get_net -of_object [get_pin $iterm]]
if { $net != "NULL" } {
  puts "Net connected to b2/u3/Z: [get_full_name $net]"
} else {
  puts "b2/u3/Z is not connected to any net."
}
```
[...]

PR #401, opened 2026-03-09
Fix for OpenSTA issue 398 and OpenROAD issue 9454 with regression

dsengupta0628:
Issue: https://github.com/parallaxsw/OpenSTA/issues/398
Instance b2 is of the type block2 which has a 2nd output pin "out2". Even if it is unconnected to any net in the top level (i.e. the "top" module), the pin b2/out2 should be still a valid object.

Potential fix per feedback in the issue.

So in `VerilogReader::makeModuleInstNetwork` I just enhanced the existing code that was for liberty cells only to include non-liberty cells too
Needed an updation in ` VerilogReader::makeInstPin` because we have already called makePin now even for non-liberty cells by the new change- so need to make sure we don't create duplicate here.

Added a regression for the issue.

### DIFF
--- a/verilog/VerilogReader.cc
+++ b/verilog/VerilogReader.cc
@@ -1825,11 +1825,12 @@ VerilogReader::makeModuleInstNetwork(VerilogModuleInst *mod_inst,
     }
 
     // Make all pins so timing arcs are built and get_pins finds them.
-    ConcreteCellPortBitIterator port_iter(reinterpret_cast<const ConcreteCell*>(cell));
-    while (port_iter.hasNext()) {
-      Port *port = reinterpret_cast<Port*>(port_iter.next());
+    CellPortBitIterator *port_iter = network_->portBitIterator(cell);
+    while (port_iter->hasNext()) {
+      Port *port = port_iter->next();
       network_->makePin(inst, port, nullptr);
     }
+    delete port_iter;
     bool is_leaf = network_->isLeaf(cell);
     VerilogBindingTbl bindings(zero_net_name_, one_net_name_);
     if (mod_inst->hasPins()) {


test/regression_vars.tcl
```diff
@@ -166,6 +166,7 @@ record_public_tests {
   verilog_attribute
   verilog_specify
   verilog_write_escape
+  verilog_unconnected_dbterm
 }
```
parallaxsw, line 170: dbterm is ODB terminolgy, which is irrelevant here. use unconnected_hpin

test/verilog_unconnected_dbterm.tcl
```diff
@@ -0,0 +1,14 @@
+read_liberty ../examples/nangate45_typ.lib.gz
```
parallaxsw: asap7_small.lib is a smaller library to use for simple tests

test/verilog_unconnected_dbterm.tcl
```diff
@@ -0,0 +1,14 @@
...
+puts "Find b2/out2: [get_property [get_pins b2/out2] full_name]"
+# Check if net is connected to "b2/u3/Z" that was the b2/out2 in parent block
+set iterm [sta::find_pin "b2/u3/Z"]
+set net [get_net -of_object [get_pin $iterm]]
+if { $net != "NULL" } {
+  puts "Net connected to b2/u3/Z: [get_full_name $net]"
+} else {
+  puts "b2/u3/Z is not connected to any net."
+}
+puts "Find internal net connected to b2/out2: [get_property [get_net b2/out2] full_name]"
```
parallaxsw: use get_full name like above instead of get_property
but both are overkill. easier to just use report_net to see the pin listed than all of this
printing/if logic

verilog/VerilogReader.cc
```diff
@@ -1983,12 +1981,16 @@ VerilogReader::makeInstPin(Instance *inst,
...
-    Pin *pin = network_->makePin(inst, port, net);
-    if (!is_leaf && net) {
-      const char *port_name = network_->name(port);
-      Net *child_net = bindings->ensureNetBinding(port_name, inst, network_);
-      network_->makeTerm(pin, child_net);
-    }
+    // Ensure pin exists (may have been pre-created), then connect to parent
+    // net if present. Always create a term for the child-side net.
+    Pin *pin = network_->findPin(inst, port);
+    if (pin == nullptr)
```
parallaxsw: This cannot happen now that all of the pins are already made.

verilog/VerilogReader.cc
```diff
@@ -1824,13 +1824,11 @@ VerilogReader::makeModuleInstNetwork(VerilogModuleInst *mod_inst,
...
-    if (lib_cell) {
-      // Make all pins so timing arcs are built.
-      LibertyCellPortBitIterator port_iter(lib_cell);
-      while (port_iter.hasNext()) {
-        LibertyPort *port = port_iter.next();
-        network_->makePin(inst, reinterpret_cast<Port*>(port), nullptr);
-      }
+    // Make all pins so timing arcs are built and get_pins finds them.
+    ConcreteCellPortBitIterator port_iter(reinterpret_cast<const ConcreteCell*>(cell));
+    while (port_iter.hasNext()) {
```
parallaxsw: I think this would be better off using Network::portBitIterator so there are no casts involved.

dsengupta0628: Network::portBitIterator returns a heap-allocated CellPortBitIterator*, So I will have to use a delete on it later, would that be okay? (this is the pattern used in ConcreteNetwork::makePins)

Otherwise I can also use  std::unique_ptr<CellPortBitIterator> port_iter(network_->portBitIterator(cell))

Please let me know your preference. Thanks

parallaxsw: yes, you have to delete it either explicitly (like all the other network iterator calls you can find in OpenSTA) or use unique_ptr which I haven't adopted much because it makes the declaration lines sooooo wiiiiddddde.

dsengupta0628: Gotcha! Done.

PR #402, opened 2026-03-10
Make `wire_load_tree_` default to unknown so we can detect when it's not present in a Liberty file.

rocallahan:
This is potentially a behavior change, but I think omitting this from the Liberty file is rare. I've only seen it in some DTCD Liberty files. In those Liberty libraries, it seems to be expected that the Liberty data is valid for all `WireloadTree` types. Thus it is necessary to distinguish between "wire load tree was specified as 'balanced'" and "wire load tree was not specified".


parallaxsw:
EstimateParasitics::estimatePiElmore treats unknown as balanced,
but 
Parasitics::makeWireloadNetwork treats unknown as best_case

so you should update the second use to be consistent with the missing
tree_type

PR #408, opened 2026-03-24
sdc: Implements recursive descent parser for filter expressions

### LINKED ISSUE #399: filter_objs proc does not handle parentheses in the filter expression
https://github.com/parallaxsw/OpenSTA/blob/62cd210a8aff98cb8bf795e7b70c610faf5deebb/sdc/Sdc.tcl#L319-L350

Attempts to parse filter expressions using regex which introduces limitations in the type of expressions it can handle -- for one the handing of parentheses..

This issue is to expand the support to handle expressions such as
```
(property1 == val1) && (property2 == val2)
```
and more complex filter expressions.

Testcase:
```
puts "with"
puts [filter_objs "(property == val1) && (property == val2)" "objects" "concat" "object_type"]
puts "without"
puts [filter_objs "property == val1 && property == val2" "objects" "concat" "object_type"]
```

Expected identical output.  Instead it drops the second expression after the &&
```
with
property == val1 objects
without
property == val2 property == val1 objects
```

QuantamHD:
This PR introduces a proper recursive descent parser for OpenSTA's filter expression fixing previously reported errors with the correct evaluation of expressions containing parentheses.

Fixes https://github.com/parallaxsw/OpenSTA/issues/399


jjcherry56:
I would much rather see it in c++ than tcl, so the Silimate implementation looks better.
I also don't like that the proc names are no very specific to the application (filters).

akashlevy:
@QuantamHD Cherry just upstreamed Silimate's C++ implementation, can you double check that this also does what you were hoping?

PR #415, opened 2026-04-02
util/StringUtil.cc: feature-test std::from_chars

donn:
Support for std::{from,to}_chars isn't finalized in
libcxx as of the time of writing, see __cpp_lib_to_chars in
https://github.com/llvm/llvm-project/blob/6331bfa41ab529558ec9d645c0effb7a4146591c/libcxx/docs/FeatureTestMacroTable.rst

This patch adds a fallback using strtof. There are two
differences:
* strtof is locale-dependent
* strtof tolerates leading spaces

### DIFF
--- a/util/StringUtil.cc
+++ b/util/StringUtil.cc
@@ -28,6 +28,7 @@
 #include <cctype>
 #include <charconv>
 #include <system_error>
+#include <version>
 
 namespace sta {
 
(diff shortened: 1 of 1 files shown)


donn:
Notes:

- Despite lacking the feature flag, macOS 26 *does* support the floating-point variant of std::from_chars. However, 15.X and earlier do not and explicitly mark it as deleted.
- I ran this small script to verify that the behavior is consistent with libstdc++:
  ```
  cat <<HEREDOC > ./test.tcl
  foreach str {
      1.321e-1
      1.321e10
      1.321
      -4.15
      " 4"
      "4 "
  } {
      puts [sta::format_time \$str 8]
  }
  HEREDOC
  ./build/sta -exit -no_splash test.tcl > mac.txt
  lima ./cmake-build-debug/sta -exit -no_splash test.tcl > linux.txt
  diff linux.txt mac.txt
  ```
  
  The only behavioral inconsistency is its handling of a leading space, which std::stringstream ignores but std::from_chars flags as invalid and returns 0.

jjcherry56:
I would rather see the alternative use strtof than string streams. locale-independence is not really relevant because it is not supported anywhere in OpenSTA.

PR #420, opened 2026-04-09
Fix coverity issues

dsengupta0628:
Summary of changes- all Coverity complaints so far. I will monitor for more too. Thanks!

coverity

### DIFF
--- a/include/sta/ContainerHelpers.hh
+++ b/include/sta/ContainerHelpers.hh
@@ -437,7 +437,7 @@ void
 sort(Range& r,
      Comp comp = Comp{})
 {
-  std::sort(std::ranges::begin(r), std::ranges::end(r), comp);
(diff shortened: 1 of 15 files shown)


parallaxsw:
When I made the switch to std::string I did a bunch of research and went back and forth, but in the end I converged on mostly using std::string_view for args and not inserting std::move's all over the place. I cannot run coverity (it takes them something like 6 days to approve a request), but I pushed some changes that are an alternative to some of these changes.

dsengupta0628:
Yes majority of these are the std::move related and thanks for fixing changes in these two files. I will merge them and reupload with whatever is remaining (currently as expected I have some merge conflicts anyways).

dsengupta0628:
Most have been taken care of. Thanks

PR #421, opened 2026-04-10
fix: add typename and this-> for dependent names in Iterator.hh

oharboe:
GCC 15 and clang 20 enforce -Wtemplate-body, requiring typename before dependent type names and this-> for member accesses whose type depends on template parameters. Without these qualifiers, two-phase name lookup fails to resolve itr_ in VectorIterator, MapIterator and SetIterator.

### DIFF
--- a/include/sta/Iterator.hh
+++ b/include/sta/Iterator.hh
@@ -48,7 +48,7 @@ public:
     seq_(seq)
   {
     if (seq_)
-      itr_ = seq_->begin();
+      this->itr_ = seq_->begin();
   }
   VectorIterator(const VECTOR_TYPE &seq) :
     seq_(&seq),
@@ -56,12 +56,12 @@ public:
   {
   }
 
-  bool hasNext() override { return seq_  && itr_ != seq_->end(); }
-  OBJ_TYPE next() override { return *itr_++; }
+  bool hasNext() override { return seq_  && this->itr_ != seq_->end(); }
+  OBJ_TYPE next() override { return *this->itr_++; }
 
 protected:
   const VECTOR_TYPE *seq_;
-  VECTOR_TYPE::const_iterator itr_;
+  typename VECTOR_TYPE::const_iterator itr_;
 };
 
 template <typename MAP_TYPE, typename OBJ_TYPE>
@@ -72,7 +72,7 @@ public:
     map_(map)
   {
     if (map)
-      itr_ = map->begin();
+      this->itr_ = map->begin();
   }
   MapIterator(const MAP_TYPE &map) :
     map_(&map),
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller FYI

parallaxsw:
I compiled with clang 22 and do not see an issue.

# GCC and Clang template body enforcement
_Exported on 4/10/2026 at 09:54:42 PDT from Cursor (3.0.13)_

---

**User**

is this true?
GCC 15 and clang 20 enforce -Wtemplate-body, requiring typename before dependent type names and this-> for member accesses whose type depends on template parameters. Without these qualifiers, two-phase name lookup fails to resolve itr_ in VectorIterator, MapIterator and SetIterator.


---

**Cursor**

Let me look at the relevant iterator classes in the codebase and research the compiler warning.

The claim is **not true** for this code, for two separate reasons. Let me break down what's accurate and what's wrong.

## 1. `-Wtemplate-body` is real, but only in GCC

GCC 15 did introduce `-Wtemplate-body` — it extends template body checking to catch errors in uninstantiated templates (e.g., `this->nonexistent_member()`). However, **Clang 20 does not have a `-Wtemplate-body` flag**. Clang has always been stricter about two-phase lookup, but it uses different diagnostics.

## 2. `typename` is NOT required here — you're compiling C++20

The project uses C++20:

```624:624:CMakeLists.txt
target_compile_features(OpenSTA PUBLIC cxx_std_20)
```

C++20 (via [P0634R3](https://wg21.link/P0634R3) "Down with typename!") relaxed the `typename` requirement in contexts where only a type is valid. A **member variable declaration** is exactly such a context. So these lines are perfectly valid C++20:

```64:64:include/sta/Iterator.hh
  VECTOR_TYPE::const_iterator itr_;
```

```92:92:include/sta/Iterator.hh
  MAP_TYPE::const_iterator itr_;
```

```116:116:include/sta/Iterator.hh
  SET_TYPE::const_iterator itr_;
```

Both GCC 9+ and Clang 16+ implement P0634R3, so GCC 15 and Clang 20 both accept this without `typename`.

## 3. `this->` is NOT needed for `itr_` — it's a direct member, not inherited

The claim that two-phase lookup "fails to resolve `itr_`" is **incorrect**. The `this->` requirement applies only to members inherited from a **dependent base class**. Here, `itr_` is declared directly in each class template itself (in the `protected:` section), not in the base `Iterator<OBJ_TYPE>`. For example:

```43:65:include/sta/Iterator.hh
template <typename VECTOR_TYPE, typename OBJ_TYPE>
class VectorIterator : public Iterator<OBJ_TYPE>
{
  // ...
protected:
  const VECTOR_TYPE *seq_;
  VECTOR_TYPE::const_iterator itr_;   // declared RIGHT HERE in VectorIterator
};
```

When `itr_` is used in the constructor or `hasNext()`/`next()`, it's found by ordinary unqualified lookup within the class scope — no dependent base traversal is involved. The base class `Iterator<OBJ_TYPE>` only declares the pure virtual `hasNext()` and `next()`; it has no `itr_` member at all.

## Summary

| Claim | Verdict |
|---|---|
| GCC 15 enforces `-Wtemplate-body` | **True** — but it catches `this->` calls to non-existent members, not the issues described |
| Clang 20 enforces `-Wtemplate-body` | **False** — Clang doesn't have this flag |
| `typename` is required before `VECTOR_TYPE::const_iterator` | **False in C++20** — P0634R3 makes it optional in member declarations |
| `this->` is needed for `itr_` | **False** — `itr_` is a direct member, not inherited from a dependent base |

If you were compiling this as C++17 or earlier, the missing `typename` on lines 64, 92, and 116 would indeed be an error. But since you're targeting C++20, the code as written is correct.

oharboe:
:+1:

PR #424, opened 2026-04-14
Use fully qualified `::sta::float_inf` - Tcl9 lookup changed.

### LINKED ISSUE #422: Tcl 9: can't see `sta::float_inf` inside `report_check_types`
Calling `report_check_types` when compiled with Tcl 9 results in an issue reading the `sta::float_inf` variable.

Let's consider the following `foo.tcl`

```tcl
cat ~/foo.tcl 
proc hello {} {
  # Accessing vars from within proc
  puts "Tcl patchlevel = $::tcl_patchLevel"
  puts "sta::float_inf = $sta::float_inf"
}

hello
report_check_types
```

Calling this with a **Tcl8**-linked sta, we can see that we can access `$sta::float_inf` as expected within a proc, and calling `report_check_types` emits some error as expected

```
cat ~/foo.tcl | build-8/sta -no_splash
Tcl patchlevel = 8.5.19
sta::float_inf = 1.0000000150474662e+30
Error: 1571 No network has been linked.
```

However, compiled with **tcl 9** there is a strange error from within `report_check_types` as it can't access the variable, while our little `hello` test proc can see and access it.

```
$ cat ~/foo.tcl | build-9/sta -no_splash
Tcl patchlevel = 9.0.3
sta::float_inf = 1.0000000150474662e+30
can't read "sta::float_inf": no such variable
```

Unclear the reason to me right now.
Some different way Tcl 9 handles variables outside the current scope compared to Tcl 8.x ?
Some swig issue (using `SWIG Version 4.3.1`) ?
`proc_redirect` confusing Tcl 9 ?
Bug in Tcl 9 ?

hzeller:
To get the value of a global variable, it seems in Tcl9, this requires to fully qualify it with leading `::`.

While at it, make `float_inf` readonly, as it certainly is not meant to be modified.

Thanks to @mikesinouye for spotting the problem.

Fixes: #422

### DIFF
--- a/search/Search.tcl
+++ b/search/Search.tcl
@@ -485,12 +485,12 @@ proc_redirect report_check_types {
     }
     if { $violators } {
       set group_path_count $sta::group_path_count_max
-      set slack_min [expr -$sta::float_inf]
+      set slack_min [expr -$::sta::float_inf]
       set slack_max 0.0
     } else {
       set group_path_count 1
-      set slack_min [expr -$sta::float_inf]
-      set slack_max $sta::float_inf
+      set slack_min [expr -$::sta::float_inf]
+      set slack_max $::sta::float_inf
     }
 
     set path_ends [find_path_ends "NULL" {} "NULL" 0 \
--- a/util/Util.i
+++ b/util/Util.i
@@ -49,7 +49,7 @@ using namespace sta;
 
 %inline %{
 
-float float_inf = INF;
+const float float_inf = INF;
 
 const char *
 version()


hzeller:
With that, the example from #422 works with both, tcl-8 and tcl-9

```
$ cat ~/foo.tcl | build-8/sta -no_splash
Tcl patchlevel = 8.5.19
sta::float_inf = 1.0000000150474662e+30
Error: 1571 No network has been linked.
```

```
$ cat ~/foo.tcl | build-9/sta -no_splash
Tcl patchlevel = 9.0.3
sta::float_inf = 1.0000000150474662e+30
Error: 1571 No network has been linked.
```

jjcherry56:
I added variable commands to import the variables instead

PR #427, opened 2026-04-21
Make defineProperty take handler by const& and clean up std::moves

calewis:
If you leave them as non-const & then you can't pass rvalue lambdas in via this interface anymore.

### DIFF
--- a/search/Property.cc
+++ b/search/Property.cc
@@ -1265,70 +1265,70 @@ Properties::capacitancePropertyValue(float cap)
 
 void
 Properties::defineProperty(std::string_view property,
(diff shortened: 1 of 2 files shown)


search/Property.cc
```diff
@@ -1355,7 +1355,7 @@ void
 PropertyRegistry<TYPE>::defineProperty(std::string_view property,
                                        PropertyHandler handler)
 {
-  registry_[std::string(property)] = std::move(std::move(std::move(std::move(std::move(std::move(std::move(std::move(std::move(std::move(handler))))))))));
+  registry_[std::string(property)] = std::move(handler);
```
jjcherry56, line 1358: WTF, claude stuttering

PR #431, opened 2026-04-29
Check is_object before accessing string offset in object_type

calewis:
We have fuzz tests that were generating failures on this method.  Since i doubt `object_type` is a performance hot spot it seems worth just making sure the dereference is safe.


jjcherry56:
This was resolved in
7e68e452d James Cherry 2026-05-07 mv is_object to Util.i
which also updated all the callers to make sure the arg passed is_object first

PR #437, opened 2026-05-14
#436 fixed incorrect bit width calculation

JayPankajPatel:
## Issue
writeBusDcls() in liberty/LibertyWriter.cc computes bit_width incorrectly for ascending bus ranges (e.g. [0:10]).

## Root Cause

```cpp
// before — +1 inside abs(), wrong for ascending ranges
std::abs(dcl->from() - dcl->to() + 1)
// ascending  [0:10]:  abs(0  - 10 + 1) = abs(-9) = 9   ← WRONG
// descending [10:0]:  abs(10 - 0  + 1) = 11             ← correct

// after — +1 outside abs(), correct for both
std::abs(dcl->from() - dcl->to()) + 1
// ascending  [0:10]:  abs(0  - 10) + 1 = 11  ✓
// descending [10:0]:  abs(10 - 0)  + 1 = 11  ✓
```

## Change

One character moved — `+1` from inside `std::abs()` to outside.

Tested against OpenSTA 2.5.0.

### DIFF
--- a/liberty/LibertyWriter.cc
+++ b/liberty/LibertyWriter.cc
@@ -275,7 +275,7 @@ LibertyWriter::writeBusDcls()
     sta::print(stream_, "  type (\"{}\") {{\n", dcl->name());
     sta::print(stream_, "    base_type : array;\n");
     sta::print(stream_, "    data_type : bit;\n");
-    sta::print(stream_, "    bit_width : {};\n", std::abs(dcl->from() - dcl->to() + 1));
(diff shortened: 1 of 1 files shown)


jjcherry56:
This should include the test from the issue.
see test/verilog_write_escape.tcl for an example of how to avoid all the tcl in the test case.
the file contents is all that is necessary to verify the result

JayPankajPatel:
Sure I believe I added a test enviorment I used in #436  but I will redo it after taking a look at test/verilog_write_escape.tcl

jjcherry56:
I saw no motion on adding a test case so I added one in the private tests.

PR #439, opened 2026-05-19
fix makeConcreteParasitics leak+testcase

dsengupta0628:
### **Problem- Sta::makeConcreteParasitics map-overwrite leak:**
 
The bug is  in Sta::makeConcreteParasitics in search/Sta.cc, as parasitics_name_map_ is std::map<std::string, Parasitics*>. operator[]= replaces the raw pointer without delete, so when the key already exists, the prior Parasitics* is orphaned.

Neither Sta::deleteParasitics()  nor ~Sta() can reach it again because both iterate parasitics_name_map_.

  ```
 Parasitics *
   Sta::makeConcreteParasitics(std::string_view name,
                               std::string_view filename)
   {
     Parasitics *parasitics = new ConcreteParasitics(name, filename, this);
     parasitics_name_map_[std::string(name)] = parasitics;   // overwrites
     return parasitics;
   }
```

### **Reproducing test along with this PR:** 
test/make_concrete_parasitics_leak.tcl, two read_spef -min calls with the same scene+min_max combination hit the same map key and orphan the prior ConcreteParasitics. Crashes with ASAN run (steps followed to repro)

1.  mkdir -p build-asan && cd build-asan
2.  cmake .. -DCMAKE_BUILD_TYPE=Debug  -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address"  -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address"
3.  cmake --build . --target sta -j$(nproc)
4.  cd ../test
5.  ASAN_OPTIONS=detect_leaks=1:abort_on_error=1:disable_coredump=0 ../build-asan/sta -no_init -no_splash -exit make_concrete_parasitics_leak.tcl

  LeakSanitizer will print a direct leak whose top frames are

```
   Sta::makeConcreteParasitics  search/Sta.cc:4295   (the new ConcreteParasitics)
   Sta::readSpef                search/Sta.cc:4178
```
 plus an indirect leak chain

```
   vector<Mode*>/vector<Scene*>::operator=
   StaState::operator=          include/sta/StaState.hh:59
   StaState::StaState(StaState const*)  search/StaState.cc:62
   Parasitics::Parasitics(StaState*)    parasitics/Parasitics.cc:42
   ConcreteParasitics::ConcreteParasitics  parasitics/ConcreteParasitics.cc:747
   Sta::makeConcreteParasitics  search/Sta.cc:4295
```

### **Fix in PR**: 
If key already mapped, wipe contents ( ConcreteParasitics::clear() is already used by deleteParasitics() ) and reuse same object. Scene back-pointers stay  valid because the object never moves. Verified fix with ASAN build. But not sure how to enforce running ASAN for this testcase as part of regular regression running.

### CODE REVIEW

File: test/make_concrete_parasitics_leak.tcl [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+read_liberty asap7_small.lib.gz
+read_verilog reg1_asap7.v
+link_design top
+create_clock -name clk -period 500 {clk1 clk2 clk3}
+set_input_transition 10 {in1 in2 clk1 clk2 clk3}
+# 1st read_spef
+read_spef -min reg1_asap7.spef
+# Mid-flow: query timing, switch delay calculators.
+report_checks -path_delay min -digits 3
+set_delay_calculator arnoldi
```
jjcherry56: this is pointless. all you need is 2 read_spef calls.

search/Sta.cc
```diff
@@ -4291,7 +4291,13 @@ Parasitics *
 Sta::makeConcreteParasitics(std::string_view name,
                             std::string_view filename)
 {
-  Parasitics *parasitics = new ConcreteParasitics(name, filename, this);
+  // Reuse existing entry to avoid leaking the prior Parasitics.
+  Parasitics *parasitics = findParasitics(std::string(name));
```
jjcherry56: this does not preserve the filename

test/make_concrete_parasitics_leak.tcl
dsengupta0628: Yes- I just wanted to make it sort of a realistic flow. But will change to my original two read_spefs back to back

search/Sta.cc
dsengupta0628: Understood. Using delete instead so the filename is preserved in subsequent calls without leaks

test/make_concrete_parasitics_leak.tcl
dsengupta0628, line 1: Updated now, and the .ok file is just empty- I saw other tests in this directory do have empty .ok files, hence kept as is

search/Sta.cc
```diff
@@ -4291,8 +4291,15 @@ Parasitics *
 Sta::makeConcreteParasitics(std::string_view name,
                             std::string_view filename)
 {
+  // Free the prior entry to avoid leaking it on overwrite.
+  std::string key(name);
+  auto it = parasitics_name_map_.find(key);
+  if (it != parasitics_name_map_.end()) {
+    delete it->second;
+    parasitics_name_map_.erase(it);
```
jjcherry56, line 4299: boy do i hate c++ iterators. sooo verbose

maliberty: Not sure its better but you can one line it:
```
if (auto it = parasitics_name_map_.find(key); it != parasitics_name_map_.end()) {
```

PR #441, opened 2026-05-28
write_verilog: escaped-scalar port reference mismatch

oharboe:
For per-bit escaped-scalar ports (single-bit ports whose internal name contains brackets, with no aggregating bus dcl) — e.g. the output of OpenROAD partition-synth + repair_design — `write_verilog` emits dcl/reference forms that don't match:

- `writePortDcls` / `writeWireDcls` route through `portVerilogName` → `\foo[7] ` (escaped scalar).
- `writeInstPin` / `writeInstBusPinBit` / `writeAssigns` route through `netVerilogName` → `foo[7]` (bus-select on undeclared `foo`).

Verilator rejects the resulting netlist:

```
%Error: out.v:198120:65: Illegal bit or array select;
        type does not have a bit range, or bad dimension: data type is 'logic'
```

Track bus prefixes whose dcl is actually emitted in the current module; use `netVerilogName` (bus-select) only when a matching bus dcl exists, otherwise fall back to `portVerilogName` (escaped scalar). Cleared per `writeModule`.

The existing `verilog/test/verilog_escaped_write_*` regressions use proper bus wires/ports, so `declared_buses_` matches and they emit unchanged.

### DIFF
--- a/verilog/VerilogWriter.cc
+++ b/verilog/VerilogWriter.cc
@@ -27,6 +27,7 @@
 #include <algorithm>
 #include <cstdlib>
 #include <map>
+#include <set>
 #include <string>
 #include <string_view>
 
@@ -73,6 +74,7 @@ protected:
                           const Port *port,
                           bool &first_member);
   void writeAssigns(const Instance *inst);
+  std::string netVerilogNameInScope(std::string_view net_name);
 
   int findUnconnectedNetCount(const Instance *inst);
   int findChildNCcount(const Instance *child);
@@ -85,6 +87,7 @@ protected:
   FILE *stream_;
   Network *network_;
   int unconnected_net_index_{1};
+  std::set<std::string, std::less<>> declared_buses_;
 };
 
 void
@@ -172,6 +175,7 @@ VerilogWriter::findHierChildren(const Instance *inst,
 void
 VerilogWriter::writeModule(const Instance *inst)
 {
+  declared_buses_.clear();
   Cell *cell = network_->cell(inst);
   std::string cell_vname = cellVerilogName(std::string(network_->name(cell)));
   sta::print(stream_, "module {} (", cell_vname);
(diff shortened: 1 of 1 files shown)


oharboe:
@jjcherry56 Claude fixed a Verilator syntax error for us :facepalm: when we do verilator netlist simulation to get a saif file for OpenSTA.

jjcherry56:
Please provide a simple testcase so I can reproduce the problem

oharboe:
Untar [opensta-441-repro.tar.gz](https://github.com/user-attachments/files/28368075/opensta-441-repro.tar.gz) and run STA=bin/sta ./run.sh

```
$ STA=/fullpath/bin/sta ./run.sh 

=== write_verilog output (out.v) ===
module top (in,
    \out[0] );
 input in;
 output \out[0] ;


 BUFx2_ASAP7_75t_R u1 (.Y(out[0]),
    .A(in));
endmodule

=== verilator --lint-only (optional) ===
%Warning-IMPLICIT: out.v:7:27: Signal definition not found, creating implicitly: 'out'
    7 |  BUFx2_ASAP7_75t_R u1 (.Y(out[0]),
      |                           ^~~
                   ... For warning description see https://verilator.org/warn/IMPLICIT?v=5.032
                   ... Use "/* verilator lint_off IMPLICIT */" and lint_on around source to disable this message.
%Error: out.v:7:30: Illegal bit or array select; type does not have a bit range, or bad dimension: data type is 'logic'
                  : ... note: In instance 'top'
    7 |  BUFx2_ASAP7_75t_R u1 (.Y(out[0]),
      |                              ^
%Error: Exiting due to 1 error(s)

=== syntax check ===
FAIL: bus-select reference(s) to undeclared bus (PR #441 bug):
  out[0]
```

jjcherry56:
I think claude is confused.
The problem is in the names used to make the port/net, NOT the verilog writer.
The internal sta name should have escapes for the brackets.
I will fix the tcl  functions used in your testcase, but I believe the openroad partition code
need to be changed to use the correct names.

oharboe:
Makes sense. So it sounds like I can close this PR, we have picked this bone clean of value?

jjcherry56:
I pushed a fix for the opensta tcl side your testcase reproduces.
13cd475ff make_net/make_port bus bit without bus dcl pr441

oharboe:
:+1:

PR #443, opened 2026-06-02
Dont spin main for queue

calewis:
Replace the busy-yielding pending_task_count_ loop in DispatchQueue::finishTasks with a blocking DynamicLatch. This avoids having the main thread consume CPU cycles while waiting for dispatched tasks to complete.

The DynamicLatch implementation uses C++20 std::atomic::wait/notify_all for efficient blocking and wakeup.  It functions like [std::latch](https://en.cppreference.com/cpp/thread/latch) except you can also count up with it.

### DIFF
--- a/include/sta/DispatchQueue.hh
+++ b/include/sta/DispatchQueue.hh
@@ -1,11 +1,7 @@
-// Author Phillip Johnston
-// Licensed under CC0 1.0 Universal
-// https://github.com/embeddedartistry/embedded-resources/blob/master/examples/cpp/dispatch.cpp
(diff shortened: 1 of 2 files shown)


jjcherry56:
I couldn't find any problems in my testing. Unfortunately it didn't do anything to imrove run time, just user time.

calewis:
> I couldn't find any problems in my testing. Unfortunately it didn't do anything to imrove run time, just user time.

When threads <= cores, it's just nicer behavior. You aren't spinning a core in a busy loop. When threads > cores, either by accident or in some testing env, this actually frees up a core to do work. It also should let other processes in the background make progress. But yeah, in a situation with a quiet machine and threads <= cores, you shouldn't expect a  wall time gain.

PR #444, opened 2026-06-04
Add copy() methods to Parasitic classes.

Logikable:
We'd like to parallelize calling gateDelay on potentially the same instance/net.

### DIFF
--- a/dcalc/Arnoldi.hh
+++ b/dcalc/Arnoldi.hh
@@ -69,7 +69,10 @@ class rcmodel : public ConcreteParasitic,
                 public arnoldi1
 {
 public:
+  rcmodel();
+  rcmodel(const rcmodel &rc);
   ~rcmodel() override;
+  Parasitic *copy() const override;
   float capacitance() const override;
   PinSet unannotatedLoads(const Pin *drvr_pin,
                           const Parasitics *parasitics) const override;
--- a/dcalc/ArnoldiReduce.cc
+++ b/dcalc/ArnoldiReduce.cc
@@ -43,6 +43,51 @@ namespace sta {
 // This is legacy C-style code.
 // NOLINTBEGIN(modernize-avoid-c-style-cast, bugprone-multi-level-implicit-pointer-conversion, bugprone-implicit-widening-of-multiplication-result)
 
+rcmodel::rcmodel() :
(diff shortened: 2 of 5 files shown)


jjcherry56:
That does not make any sense. 
Why do you think you have to copy the parasitics?

tspyrou:
Perhaps gate changes require parallel reduction but shouldnt require the detailed parasitics to be copied. Some explanation of the use would help.

Logikable:
When evaluating a resize candidate I'd like to compute the delay of said candidate. Normally we modify the Parasitic temporarily as per [1], which is then used in [2]. Doing this isn't thread-safe, though.

You're right that we don't need the detailed parasitics to be copied. I'd be happy with a solution where the ConcretePiElmore could be copied, but the header for that class is private.

We could restrict each instance to one thread, thereby not needing a thread-safe Parasitic mutation. I've run into long-tails where most of the threads are done, and the wait-time is dominated by a single instance's resize candidates.

I'm open to hearing alternatives.

[1] https://github.com/The-OpenROAD-Project/OpenROAD/blob/8df74aa8d0384f76a742046312f0125885f258c3/src/rsz/src/DelayEstimator.cc#L622
[2] https://github.com/The-OpenROAD-Project/OpenROAD/blob/8df74aa8d0384f76a742046312f0125885f258c3/src/rsz/src/DelayEstimator.cc#L645

jjcherry56:
Copying looks like really a work around for calling the constructor.

It seems like it would make more sense to just include ConcreteParasiticsPvt.hh (yes, it is a private header) and then call the ConcretePiElmore constructor. Or ask for a function to
create a temporary. But I still see no point in copying.
The temporary parasitic does not live in the parasitics database and is a legitimate argument
to the delay calculator.

tspyrou:
Can you make ConcreteParasiticsPvt.hh public or at thread up this constructor?

jjcherry56:
The problem is the code in question assumes the parasitic needed by the delay calculator is a pi elmore reduced mode. It will only work with one delay calculator. Not prima (ccs), not arnoldi, not LVF. 

So change the include paths in the cmake file to bypass the deliberate hiding of the parasitic/delay calculator details. Make it clear that the code is going around the APIs.

PR #445, opened 2026-06-09
Skip CRPR machinery when no clock is propagated (complete fb5d87bd)

oharboe:
## Summary
Make `crprActive()` additionally require that the SDC has at least one propagated clock. With ideal clocks CRPR can produce no correction, so this skips the CRPR search/CheckCrpr machinery entirely in that case.

This completes the intent of fb5d87bd ("no crpr for ideal clocks"): that commit added `TagGroupBldr::hasPropagatedClk()` but only consulted it when skipping `pruneCrprArrivals()`. The gate that actually enables CRPR, `StaState::crprActive()`, still keyed only off `analysisType()==ocv && crprEnabled()`.

## Motivation
In pre-CTS placement-stage timing, OCV is active and `crpr_enabled` is on, but clocks are ideal — propagation is applied after CTS. On high-reconvergence clock structures the CRPR path can consume very large amounts of memory while contributing nothing, since the correction is already zero for ideal clocks. Setting `set_crpr_enabled 0` avoids it, which confirms the cost is pure overhead in this case.

This matters for automated design-space exploration, which runs STA at placement to prune configurations cheaply — including pathological ones that we specifically want to reject *before* paying for CTS and routing. A configuration that blows up CRPR memory at placement can OOM-kill the process and derail the whole sweep, instead of simply being scored and pruned. Skipping the inert CRPR work at placement keeps that early pruning bounded and robust, while full CRPR still applies post-CTS where it has effect.

## Why this is result-neutral
`CheckCrpr::checkCrpr1` / `outputDelayCrpr1` already return 0 when clocks are ideal (`isPropagated()` guards), and `ClkInfo` already stores a null crpr clock path for unpropagated clocks. This change only avoids building machinery to feed a no-op. Once any clock is propagated (post-CTS / sign-off), behavior is unchanged.

## Change
- `Sdc::hasPropagatedClock()` — cached predicate (true if any clock or clock pin is propagated), invalidated on set/remove propagated clock, clock creation, and clock removal.
- `StaState::crprActive()` — additionally require `sdc->hasPropagatedClock()`.

## Validation
- All-ideal-clock pre-CTS report: peak RSS drops to the `set_crpr_enabled 0` level; reported slacks/WNS/TNS unchanged.
- Post-CTS (propagated) report: identical results and memory to before.
- Existing CRPR regressions unchanged.

### DIFF
--- a/include/sta/Sdc.hh
+++ b/include/sta/Sdc.hh
@@ -409,6 +409,8 @@ public:
   void setPropagatedClock(Pin *pin);
   void removePropagatedClock(Pin *pin);
   bool isPropagatedClock(const Pin *pin) const;
+  // True if any clock or clock pin is propagated. CRPR is inert otherwise.
+  bool hasPropagatedClock() const;
   void setClockSlew(Clock *clk,
                     const RiseFallBoth *rf,
                     const MinMaxAll *min_max,
@@ -1306,6 +1308,8 @@ protected:
   ClkHpinDisables clk_hpin_disables_;
   bool clk_hpin_disables_valid_;
   PinSet propagated_clk_pins_;
+  mutable bool has_propagated_clk_{false};
+  mutable bool has_propagated_clk_valid_{false};
   ClockLatencies clk_latencies_;
   PinSet clk_latency_pins_;
   EdgeClockLatencyMap edge_clk_latency_map_;
--- a/sdc/Sdc.cc
+++ b/sdc/Sdc.cc
@@ -167,6 +167,7 @@ Sdc::clear()
 {
   deleteConstraints();
   propagated_clk_pins_.clear();
+  has_propagated_clk_valid_ = false;
   clocks_.clear();
   clock_name_map_.clear();
   clock_pin_map_.clear();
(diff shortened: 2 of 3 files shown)


oharboe:
Tested this on the workload that motivated it — **it doesn't fix the OOM.** My pre-CTS placement report runs with the clock already propagated (ideal SDC clock, but the ORFS load path propagates it before `find_timing`), so `hasPropagatedClock()` is true and CRPR runs and OOMs as before. `unset_propagated_clock` / `set_crpr_enabled 0` is what actually drops it ~62 GB → ~8 GB.

Still worth landing as a **consistency** fix, not a memory one: it makes `crprActive()` agree with the `hasPropagatedClk()` predicate fb5d87bd already added for `pruneCrprArrivals()`, and it's result-neutral (CRPR already returns 0 for ideal clocks — this just skips building the machinery). The real 62 GB blow-up is a separate CRPR tag-scaling problem.

oharboe:
@jjcherry56 Claude found this while hunting down another problem https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/4286

Please give the code changes a quick once over to see if the consistency fix is real or imagined and if not, close the PR.

jjcherry56:
Well it looks "correct" but I would not bother with the valid flag and just recompute the value when one of the inputs changes.

I don't know why you don't just set the sta_crpr_enabled variable if you want to disable crpr. 
That does not involve changing the SDC.

oharboe:
Thank you for checking.

This was one of many investigations when servers were running out of memory, set_crpr_enabled caused the server to run out of memory.

The actual fix is here, merged.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/4286

Issue #446, opened 2026-06-09
Statistical Timing Analysis doc typos


jeffng-or:
Some minor typos in the doc

Page 3:
- "supports statistical timing .anallysis" -> "analysis"
- "uses a probaility distribution" -> "probability"
- "slew ranther than" > "rather"

Page 4:
- "with the sta_pocv_mode variaable" -> "variable"
- "Use the variance field" -> "variation"
- "design with statisical timing" -> "statistical"

Page 5
- In the report_checks command, "variation" is in the fields list twice. Prob a no-op, but wasn't sure you wanted it listed twice

dsengupta0628:
This issue is already resolved with the above PR 447 and will be available in OpenROAD/OpenSTA too once my PR is merged there (pulling latest from upstream OpenSTA). Can be closed

jeffng-or:
confirmed. Thanks!

PR #447, opened 2026-06-09
Address a few typos in docs related to SSTA

dsengupta0628:
Addressing a few SSTA doc typos from https://github.com/parallaxsw/OpenSTA/issues/446


dsengupta0628:
couple of caveats/findings:

- PDF was regenerated by my LibreOffice so likely large binary diff vs upstream's generator. If you prefer regenerating, I will drop OpenSTA.pdf from the commit. Please let me know
- I found that the variation field is actually not advertised in Search.tcl:336 help string- not sure if this was intentional.  I have not touched it, just letting you know.

jjcherry56:
The binary file issue is an inherent problem with git that there is nothing we can do about.

Please update the help string to include the variation field.

Unfortunately I too updated the doc. So please coordinate before tackling issues.

dsengupta0628:
> The binary file issue is an inherent problem with git that there is nothing we can do about.
> 
> Please update the help string to include the variation field.
> 
> Unfortunately I too updated the doc. So please coordinate before tackling issues.

Thank you. I have updated the help string now. I actually sync-d to latest parallaxsw/OpenSTA in my fork before making the change when the issue 446 came my way. But I agree, I should have sent you an email before to reduce the duplicated effort. Promise for next time!

PR #448, opened 2026-06-09
Add native gzip compression support to write_verilog

calewis:
Have `write_verilog` detect if the user requested `.gz` and if so automatically write a gzipped file.

### DIFF
--- a/test/verilog_write_gzip.tcl
+++ b/test/verilog_write_gzip.tcl
@@ -1,0 +1,8 @@
+# Check that write_verilog supports gzip compression natively if .gz extension is provided.
+source helpers.tcl
+read_liberty asap7_small.lib.gz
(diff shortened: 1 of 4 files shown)


test/verilog_write_gzip.tcl
```diff
@@ -0,0 +1,9 @@
+# Check that write_verilog supports gzip compression natively if .gz extension is provided.
+source helpers.tcl
+read_liberty gf180mcu_sram.lib.gz
+read_liberty asap7_small.lib.gz
+read_verilog verilog_write_escape.v
```
jjcherry56: use reg1_asap7.v instead

calewis: Done

PR #452, opened 2026-06-19
 Fix lost set_input_delay -reference_pin timing after graph rebuild

dsengupta0628:
**Problem** 

set_input_delay -reference_pin models the input arrival as a graph edge ref_pin --> input pin (introduced in commit 740fa7d1, port delay ref pin timing role). 

Sdc::ensureInputDelayRefPinEdges() builds the edge during levelization and sets a per-InputDelay guard **PortDelay::ref_pin_edges_exist_** to avoid duplicate edges on repeated levelization of the same graph.

But the guard is never reset when the graph is torn down. Sta::clearNonSdc() (reached via networkChangedNonSdc() - the path OpenROAD's resizer uses after netlist edits) deletes the graph, freeing the ref_pin  edge, but keeps the SDC. 
On the next rebuild the guard is still set, so ensureInputDelayRefPinEdges() skips edge re-creation. The input port loses its only arrival source and silently reports "No paths  found"

  **Fix**

Reset the guard flags at graph teardown so the edges are rebuilt with the new graph. New Sdc::inputDelayRefPinEdgesInvalid() clears the per-InputDelay flag (guarded by have_input_delay_ref_pins_), called from the clearNonSdc() modes loop just before delete graph_. The dedup guard is retained for the same-graph repeated-levelize case.

**Test**
 test/input_delay_ref_pin_rebuild.tcl: set_input_delay -reference_pin on in1, plain input delay on in2 (control), then calling sta::network_changed_non_sdc and re-report. Without the fix in1 reports "No paths found" after this call. Above fix resolves this

### DIFF
--- a/include/sta/Sdc.hh
+++ b/include/sta/Sdc.hh
@@ -577,6 +577,7 @@ public:
                         const RiseFall *clk_rf,
                         const MinMaxAll *min_max);
   void ensureInputDelayRefPinEdges();
+  void inputDelayRefPinEdgesInvalid();
   void setOutputDelay(const Pin *pin,
                       const RiseFallBoth *rf,
                       const Clock *clk,
--- a/sdc/Sdc.cc
+++ b/sdc/Sdc.cc
@@ -2840,6 +2840,18 @@ Sdc::ensureInputDelayRefPinEdges()
   }
 }
 
+// The ref_pin edges are owned by the graph. When the graph is deleted the
+// edges go with it, so the existence flags must be reset to force them to be
+// rebuilt by ensureInputDelayRefPinEdges() with the new graph.
+void
+Sdc::inputDelayRefPinEdgesInvalid()
+{
+  if (have_input_delay_ref_pins_) {
+    for (InputDelay *input_delay : input_delays_)
+      input_delay->setRefPinEdgesExist(false);
+  }
+}
+
 ////////////////////////////////////////////////////////////////
 
 void
--- a/search/Sta.cc
+++ b/search/Sta.cc
@@ -544,6 +544,8 @@ Sta::clearNonSdc()
   for (Mode *mode : modes_) {
     mode->clkNetwork()->clkPinsInvalid();
     mode->sim()->clear();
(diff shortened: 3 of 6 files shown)


parallaxsw:
You caught me.
I am removing places that insert into the bfs iterator while it is running
because they break kahn's bfs.
I know there are issues with this but let them slide because I have never
seen -reference_pin used in a design. Was it some AI generated test that
found this?

dsengupta0628:
Hi James. When I was pulling latest changes from upstream on 6/18 https://github.com/The-OpenROAD-Project/OpenSTA/pull/372 Gemini gave some code review, most of which was nonsense. 

I did not change any of your code- but I still fed Claude with Gemini's feedback regardless- and it found this error scenario. I think it found this purely because OpenSTA does support -reference_pin (I suppose for backward compatibility with older tools?)- though a similar effect could be achieved with generated clock instead of using this -reference_pin, and we would not have this error at all. And yes, I haven't seen SDCs either providing input delay like this- but for the sake of completion I thought of adding this change- upto you- can close the PR too if it seems redundant with modern SDCs.

 The test was generated by Claude using the existing libs and test infra within OpenSTA.   

Test is fairly simple though (I added in this PR)

```
read_liberty asap7_small.lib.gz
read_verilog reg1_asap7.v
link_design top
create_clock -name clk -period 500 {clk1 clk2 clk3}
set_input_delay -clock clk 100 -reference_pin r2/CLK [get_ports in1]
set_input_delay -clock clk 100 [get_ports in2]
set_input_transition 10 {in1 in2 clk1 clk2 clk3}
set_output_delay -clock clk 1 [get_ports out]

puts "=== before rebuild: in1 (ref_pin) ==="
report_checks -from in1 -path_delay max
puts "=== before rebuild: in2 (control) ==="
report_checks -from in2 -path_delay max

sta::network_changed_non_sdc

puts "=== after rebuild: in1 (ref_pin) ==="
report_checks -from in1 -path_delay max
puts "=== after rebuild: in2 (control) ==="
report_checks -from in2 -path_delay max
```

PR #453, opened 2026-06-24
write_verilog bus bit net ref without bus dcl

oharboe:
`write_verilog` emits bus-select `foo[7]` for a net reference whose name parses
as a bus member, even when the module declares no `[hi:lo] foo` bus. The
reference names an undeclared bus, so Verilator implicitly creates a 1-bit
`logic foo` and rejects the bit select.

Write side of #441: that PR escapes bracket names in `SdcNetwork::makeNet/makePort`,
but a `Network` that hands the writer a raw, unescaped bus-bit name bypasses it —
e.g. OpenROAD's `dbNetwork`, whose names come straight from the db.

Fix: track the bus prefixes actually declared by `writePortDcls`/`writeWireDcls`
and emit bus-select only when a matching bus dcl exists; otherwise fall back to
the escaped-identifier form (`portVerilogName`).

Repro (OpenROAD, which uses `dbNetwork`): a single-bit net named `foo[7]`
connected to an instance pin —

```tcl
make_port {foo[7]} input
connect_pin {foo[7]} b0/A
write_verilog out.v
```

Before / after at the instance pin (port dcl is `input \foo[7] ;` either way):

```verilog
 snl_bufx1 b0 (.Z(out), .A(foo[7]));    // before: undeclared bus -> verilator ILLEGAL
 snl_bufx1 b0 (.Z(out), .A(\foo[7] ));  // after:  matches the port dcl, lints clean
```

Existing verilog regression tests are unchanged.

### DIFF
--- a/verilog/VerilogWriter.cc
+++ b/verilog/VerilogWriter.cc
@@ -27,6 +27,7 @@
 #include <algorithm>
 #include <cstdlib>
 #include <map>
+#include <set>
 #include <string>
 #include <string_view>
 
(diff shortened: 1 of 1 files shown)


oharboe:
@jjcherry56 Claude fixed a verilator saif compilation error for us on a netlist by OpenROAD... Hopefully Claude is making sense here... :crossed_fingers:

jjcherry56:
I thought we already went over this.
The problem is in the upstream name.
foo[7] should be foo\[7\] if it not not doing a bus select.

oharboe:
> I thought we already went over this. The problem is in the upstream name. foo[7] should be foo[7] if it not not doing a bus select.

To quote Oddball in Kelly's Heroes: I only drive em, I dont know what make em work.

@maliberty I will ask Claude to articulate an OpenROAD fix  If that doesnt do it, I have to arrange a confidential testcase.

PR #454, opened 2026-06-26
make mode/scene pass as tcl objects

dsengupta0628:
Scene/Mode previously crossed the Tcl boundary as bare name strings. 
Now they pass as real SWIG handles (_..._p_Scene / _..._p_Mode), like Instance*/Clock*, so object_type dispatch works and get_scenes/get_modes return objects.

Changes:
  - search/Search.i: add Scene/Mode SWIG type stubs.
  - tcl/StaTclTypes.i: Scene*/Mode*/SceneSeq/ModeSeq in+out typemaps emit/accept handles (nullptr still --> "NULL").
  - tcl/Sta.tcl: set_scene, get_scenes -modes resolve names-->handles.
  - tcl/CmdArgs.tcl: find_scenes accepts both objects and names.
  - sdc/Sdc.tcl: read_sdc/write_sdc use cmd_mode_name instead of [cmd_mode] (handle was being used where a name string is required).

  Verified: builds clean; name and handle inputs both work; report_checks -scenes, set_scene, get_scenes -modes, mcmm3 read_sdc -mode, write_sdc all pass. (I enhanced mcmm3 to check this also)


examples/mcmm3.tcl
```diff
@@ -16,3 +16,22 @@ define_scene scene2 -mode mode2 -liberty asap7_small_ss -spef reg1_ss
 report_checks -scenes scene1
 report_checks -scenes scene2
 report_checks -group_path_count 4
+
```
jjcherry56: examples are for users.
this does not look like an example of a user script.
make a test called get_scenes instead.
use report_object_names instead of length

tcl/CmdArgs.tcl
```diff
@@ -536,14 +536,23 @@ proc parse_scenes_or_all { keys_var } {
   }
 }
 
-proc find_scenes { scene_names } {
+proc find_scenes { scene_args } {
```
jjcherry56: there is only one arg, so it should not be called args.
there are multiple scenes, so it should be scenes_arg

sdc/Sdc.tcl
```diff
@@ -69,7 +69,7 @@ proc write_sdc { args } {
     flags {-map_hpins -compatible -gzip -no_timestamp}
   check_argc_eq1 "write_sdc" $args
 
-  set mode [cmd_mode]
+  set mode [cmd_mode_name]
   if { [info exists keys(-mode)] } {
```
jjcherry56, line 73: this minimizes the changes but that isn't the right thing to optimize.
change it to work with the mode instead of mode_name

sdc/Sdc.tcl
```diff
@@ -45,7 +45,7 @@ proc_redirect read_sdc {
 
   if { [info exists keys(-mode)] } {
     set mode_name $keys(-mode)
-    set prev_mode [cmd_mode]
+    set prev_mode [cmd_mode_name]
```
jjcherry56: change it to work with the mode instead of mode_name

tcl/StaTclTypes.i
```diff
@@ -1189,106 +1189,70 @@ using namespace sta;
...
+    $1 = reinterpret_cast<Mode*>(obj);
+  }
+}
+
 %typemap(out) Mode* {
-  const Mode *mode = $1;
-  if (mode)
-    Tcl_SetResult(interp, const_cast<char*>($1->name().c_str()), TCL_VOLATILE);
+  Mode *mode = $1;
+  if (mode) {
```
jjcherry56, line 1209: there isn't any reason to support a null mode going out

tcl/StaTclTypes.i
```diff
@@ -1189,106 +1189,70 @@ using namespace sta;
...
   }
 }
 
 %typemap(out) Scene* {
-  const Scene *scene = $1;
-  if (scene)
-    Tcl_SetResult(interp, const_cast<char*>($1->name().c_str()), TCL_VOLATILE);
+  Scene *scene = $1;
+  if (scene) {
+    Tcl_Obj *obj = SWIG_NewInstanceObj(scene, SWIGTYPE_p_Scene, false);
```
jjcherry56, line 1243: there isn't any reason to support a null scene going out

examples/mcmm3.tcl
dsengupta0628: Hmm... Looks like for Scene/Mode, I need name-property support or report_object_names errors. Adding it

tcl/StaTclTypes.i
dsengupta0628: findMode/findScene return findStringKey/findKey, which yield nullptr on a missing name. The find_scene/find_mode Tcl wrappers return Scene*/Mode*, so a null does go out  through those typemaps.

To remove the null-out path, the prerequisite is reworking find_scene/find_mode (and the ~8 callers) to not use the "NULL"-sentinel pattern.

sdc/Sdc.tcl
dsengupta0628: This required a big change- please check if this makes sense now...

include/sta/Sta.hh
```diff
@@ -144,6 +144,9 @@ public:
   Mode *cmdMode() const { return cmd_mode_; }
   const std::string &cmdModeName();
   void setCmdMode(std::string_view mode_name);
```
dsengupta0628, line 146: I could not remove void setCmdMode(std::string_view mode_name);                                                                                                              because it's still live- unlike the writeSdc name overload, setCmdMode(name) has a real caller: the set_mode Tcl command.

Also setCmdMode(name) = setCmdMode(makeMode(name)) : now does create-or-find by name, then switch. That's exactly the name-based entry point set_mode  needs; only the internal restore/flip in read_sdc uses the object form (set_cmd_mode_obj --> setCmdMode(Mode*)). So both overloads are used.

jjcherry56: Ya, set_mode is overloaded to create modes so it can't take a mode instance as an arg.

sdc/Sdc.tcl
```diff
@@ -44,15 +44,15 @@ proc_redirect read_sdc {
...
 
   if { [info exists keys(-mode)] } {
-    set mode_name $keys(-mode)
+    # Create the target mode first so saving the previous mode below can never
+    # capture the throwaway default mode (which make_mode may delete).
+    set mode [make_mode $keys(-mode)]
     set prev_mode [cmd_mode]
     try {
-      set_cmd_mode $mode_name
+      set_cmd_mode_obj $mode
```
jjcherry56: yucky function name. go back to using the name i guess

sdc/Sdc.tcl
jjcherry56: does not really make sense now that i see it.

search/Sta.cc
```diff
@@ -562,44 +562,60 @@ Sta::cmdSdc() const
   return cmd_mode_->sdc();
 }
 
-void
-Sta::setCmdMode(std::string_view mode_name)
+Mode *
```
jjcherry56: undo

sdc/Sdc.tcl
dsengupta0628: Just to make sure- we go back to just the shorter change with modifying the Sdc.tcl, right? so go back set mode [cmd_mode_name] here?
Will thankfully cut down 60% of the changes in this PR!

sdc/Sdc.tcl
dsengupta0628: This too.Will bring back set prev_mode [cmd_mode_name], right?

sdc/Sdc.i
```diff
@@ -94,15 +95,18 @@ private:
...
               int digits,
               bool gzip,
               bool no_timestamp)
 {
   Sta *sta = Sta::sta();
-  sta->writeSdc(filename, mode_name, leaf, compatible, digits, gzip, no_timestamp);
+  if (mode)
+    sta->writeSdc(mode->sdc(), filename, leaf, compatible, digits, gzip, no_timestamp);
+  else
+    sta->report()->warn(1561, "mode not found.");
```
jjcherry56: I think I liked it better in the original form. This just pushes the warning up a level.
I think I was trying to push as much code as possible out of tcl by putting it in
Sta::writeSdc

dsengupta0628:
I reverted to the original set prev_mode [cmd_mode_name] usage that I had to make instead of set prev_mode [cmd_mode] that needed the whole cascade of changes relating to makeMode etc.

I kept the name-property support for Scene/Mode to support report_object_names (which is unrelated to above)

search/Property.cc
```diff
@@ -1196,6 +1197,32 @@ Properties::getProperty(const Clock *clk,
...
+    throw PropertyUnknown("scene", property);
+}
+
+////////////////////////////////////////////////////////////////
+
+PropertyValue
+Properties::getProperty(const Mode *mode,
+                        std::string_view property)
+{
+  if (property == "name"
```
jjcherry56, line 1217: this sure makes these properties look worthless

dsengupta0628: The property accessor is supposed to be a generic window into an object's attributes. Unfortunately, when the only attribute is the name (already reachable via get_name), I agree the whole mechanism is dead weight for these types..

If we had properties for each such as:
  - Scene: mode, liberty libs, parasitics/spef name
  - Mode: its scenes.

Once I finish the "-filter" implementation, then get_property with or without -filter can do something the name can't - e.g. get_scenes -filter {mode == mode1}. Until then it is indeed worthless I agree

PR #456, opened 2026-06-30
Add property-based -filter to get_scenes / get_modes, and make Scene/Mode properties extensible

dsengupta0628:
**Summary**

  Adds a -filter <expr> option to get_scenes and get_modes, matching the existing get_clocks/get_cells/etc. commands, and makes Scene/Mode properties extensible at runtime via the property registry (as Clock and the other object types already are).
  Filtering reuses the shared filterObjects<T> mechanism - no new filtering path.

Two commits specific to 2 features:

**(1)   -filter support**

  - include/sta/FilterObjects.hh, sdc/FilterObjects.cc - filterScenes / filterModes, each a filterObjects<Scene> / filterObjects<Mode> instantiation with a name-based sort comparator (mirrors filterClocks).
  - sdc/Sdc.i - filter_scenes / filter_modes SWIG wrappers.
  - tcl/StaTclTypes.i - %typemap(in) for SceneSeq* / ModeSeq* (pointer input via tclListSeqPtr); only by-value-in and out typemaps existed, and the filter_*(const char*, Seq*) signature requires the pointer-in form.
  - tcl/Sta.tcl - -filter key on both procs, applied after object resolution. get_modes reworked to parse_key_args + check_argc_eq0or1, defaulting the pattern to * when no arg is given (previously "", which matched nothing).

**(2)  Extensible Scene/Mode properties**

  - include/sta/Property.hh - defineProperty overloads for const Scene* / const Mode*; registry_scene_ / registry_mode_ members.
  - search/Property.cc - getProperty(const Scene*/Mode*) now falls back to the registry before throwing PropertyUnknown; defineProperty(Scene*/Mode*) bodies delegate to the registries.

 Lookup order now matches Clock: built-in name/full_name --> registry --> PropertyUnknown.

 **Result**

  Scene/Mode currently expose only name/full_name as built-ins, but a host (e.g. OpenROAD) can now register custom  Scene/Mode properties via defineProperty, and get_scenes -filter / get_modes -filter query them automatically - the filter layer needs no  per-property changes (unless a property introduces an entirely new PropertyValue::Type, which requires a case in the defined()/undefined() switch, as for every object type).

 **Tests**

  - test/get_scenes.tcl / .ok - added -filter coverage: scenes == / =~ / !=, modes == / =~.

### DIFF
--- a/include/sta/Property.hh
+++ b/include/sta/Property.hh
@@ -124,6 +124,10 @@ public:
                       const PropertyRegistry<const Net *>::PropertyHandler &handler);
   void defineProperty(std::string_view property,
                       const PropertyRegistry<const Clock *>::PropertyHandler &handler);
+  void defineProperty(std::string_view property,
+                      const PropertyRegistry<const Scene *>::PropertyHandler &handler);
+  void defineProperty(std::string_view property,
+                      const PropertyRegistry<const Mode *>::PropertyHandler &handler);
 
 protected:
   PropertyValue portSlew(const Port *port,
@@ -160,6 +164,8 @@ protected:
   PropertyRegistry<const Pin*> registry_pin_;
   PropertyRegistry<const Net*> registry_net_;
   PropertyRegistry<const Clock*> registry_clock_;
+  PropertyRegistry<const Scene*> registry_scene_;
+  PropertyRegistry<const Mode*> registry_mode_;
 
   Sta *sta_;
 };
--- a/include/sta/Property.hh
+++ b/include/sta/Property.hh
@@ -159,6 +159,22 @@ private:
   const Unit *unit_;
 };
 
+// Key for user-defined property values: the object instance (nullptr for
+// the property's type default), its object type name and the property name.
+class UserPropertyKey
+{
+public:
+  UserPropertyKey(const void *object,
+                  std::string_view object_type,
+                  std::string_view property);
+  bool operator<(const UserPropertyKey &key) const;
+
+private:
+  const void *object_;
+  std::string object_type_;
+  std::string property_;
+};
+
 class Properties
 {
 public:
@@ -227,19 +243,17 @@ public:
   void defineProperty(std::string_view property,
                       const PropertyRegistry<const Mode *>::PropertyHandler &handler);
 
-  // User-defined, per-object mutable properties. defineSceneProperty registers
-  // a scene property of the given type ("bool", "float" or "string") with a
-  // typed default; setSceneProperty overrides the value on one scene. The
+  // User-defined, per-object mutable properties. defineUserProperty registers
+  // a property of the given value type ("bool", "float" or "string") with a
+  // typed default; setUserProperty overrides the value on one object. The
   // property is read through the same registry path as every other property so
-  // get_property / get_scenes -filter work unchanged.
-  void defineSceneProperty(std::string_view property,
-                           std::string_view type);
-  void setSceneProperty(const Scene *scene,
-                        std::string_view property,
-                        std::string_view value);
-  void defineModeProperty(std::string_view property,
-                          std::string_view type);
-  void setModeProperty(const Mode *mode,
+  // get_property / get_* -filter work unchanged.
+  template<class TYPE>
+  void defineUserProperty(std::string_view object_type,
+                          std::string_view property,
+                          std::string_view value_type);
+  void setUserProperty(const void *object,
+                       std::string_view object_type,
                        std::string_view property,
                        std::string_view value);
 
@@ -284,11 +298,9 @@ protected:
   PropertyRegistry<const Scene*> registry_scene_;
   PropertyRegistry<const Mode*> registry_mode_;
 
-  // User-defined property values, keyed by object then property name.
-  std::map<const Scene*, std::map<std::string, PropertyValue, std::less<>>> scene_user_props_;
-  std::map<std::string, PropertyValue, std::less<>> scene_user_prop_defaults_;
-  std::map<const Mode*, std::map<std::string, PropertyValue, std::less<>>> mode_user_props_;
-  std::map<std::string, PropertyValue, std::less<>> mode_user_prop_defaults_;
+  // User-defined property values for all object types; type defaults seeded
+  // by defineUserProperty are stored with a null object.
+  std::map<UserPropertyKey, PropertyValue> user_prop_values_;
 
   Sta *sta_;
 };
--- a/include/sta/Property.hh
+++ b/include/sta/Property.hh
@@ -28,6 +28,7 @@
 #include <map>
 #include <string>
 #include <string_view>
+#include <utility>
 
 #include "LibertyClass.hh"
 #include "NetworkClass.hh"
@@ -159,19 +160,17 @@ private:
   const Unit *unit_;
 };
 
-// Key for user-defined property values: the object instance (nullptr for
-// the property's type record), its object type name and the property name.
+// Key for user-defined property values: the object instance and the
+// property name.
 class PropertyKey
 {
 public:
   PropertyKey(const void *object,
-              std::string_view object_type,
               std::string_view property);
   bool operator<(const PropertyKey &key) const;
 
 private:
   const void *object_;
-  std::string object_type_;
   std::string property_;
 };
 
@@ -282,7 +281,7 @@ protected:
   PropertyValue edgeDelay(Edge *edge,
                           const RiseFall *rf,
                           const MinMax *min_max);
-  PropertyValue propertyTypeValue(std::string_view type);
+  PropertyValue::Type propertyType(std::string_view type);
   PropertyValue coercePropertyValue(PropertyValue::Type type,
                                     std::string_view value);
 
@@ -299,8 +298,10 @@ protected:
   PropertyRegistry<const Scene*> registry_scene_;
   PropertyRegistry<const Mode*> registry_mode_;
 
-  // User-defined property values for all object types; the type record
-  // seeded by defineProperty is stored with a null object.
+  // Value types of user-defined properties keyed by object type name and
+  // property name.
+  std::map<std::pair<std::string, std::string>, PropertyValue::Type> prop_types_;
+  // User-defined property values.
   std::map<PropertyKey, PropertyValue> prop_values_;
 
   Sta *sta_;
--- a/include/sta/Property.hh
+++ b/include/sta/Property.hh
@@ -161,13 +161,13 @@ private:
 
 // Key for user-defined property values: the object instance (nullptr for
 // the property's type default), its object type name and the property name.
-class UserPropertyKey
+class PropertyKey
 {
 public:
-  UserPropertyKey(const void *object,
-                  std::string_view object_type,
-                  std::string_view property);
-  bool operator<(const UserPropertyKey &key) const;
+  PropertyKey(const void *object,
+              std::string_view object_type,
+              std::string_view property);
+  bool operator<(const PropertyKey &key) const;
 
 private:
   const void *object_;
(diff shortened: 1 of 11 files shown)


dsengupta0628:
My 3rd commit adds define_user_property and set_user_property so users can attach custom, per-object mutable properties to Scene and Mode objects, then filter on them.

  define_user_property -object_type scene -type bool active
  set_user_property [get_scenes scene1] active true
  get_scenes -filter {active == true}     ;# -> scene1
  get_property scene1 active              ;# -> 1

  Supports bool, float, and string property types.

  **How it works:**

  Values are stored in a per-object map on the Properties class, bridged into the existing per-type PropertyRegistry: define_user_property registers a handler (like any built-in property) that does the per-object lookup, with a typed default seeded at define time. So get_property and get_scenes/get_modes -filter work through the unchanged property path - no separate  filtering code. Unset objects return the type default (e.g. bool --> false), so -filter {active == true} matches only objects explicitly set.

  **Changes**

  - include/sta/Property.hh : define/setSceneProperty + define/setModeProperty decls; per-object value stores + type-default maps. (PropertyValue class moved above Properties so the maps can
  hold it by value.)
  - search/Property.cc : define/set implementations; userPropertyDefault (type --> typed default) and coerceUserValue (string --> typed value, reusing stringFloat).
  - search/Property.i : SWIG wrappers.
  - tcl/Property.tcl : define_user_property / set_user_property commands.
  - test/get_scenes.tcl / .ok : coverage for a bool active property: define, set, default fallback, -filter (explicit, shorthand, NOT), and get_property.

  **Notes**

  - Additive only; no existing command or argument changed.
  - Property set is per type (scene, mode), mirroring the codebase's per-type property convention.

include/sta/Property.hh
```diff
@@ -160,106 +281,16 @@ protected:
...
-
-  // Copy assignment.
-  PropertyValue &operator=(const PropertyValue &);
-  // Move assignment.
-  PropertyValue &operator=(PropertyValue &&) noexcept;
-
-private:
-  void destroyActive();
+  // User-defined property values, keyed by object then property name.
+  std::map<const Scene*, std::map<std::string, PropertyValue, std::less<>>> scene_user_props_;
```
jjcherry56: The double maps seem really messy to me.
It is going to get very verbose when there are user properties on any object type.
How about defining a object/object_type/property_name struct or class and use one map
for all user defined property values? Templates may be useful.

include/sta/Property.hh
```diff
@@ -160,106 +281,16 @@ protected:
...
+  std::map<std::string, PropertyValue, std::less<>> scene_user_prop_defaults_;
+  std::map<const Mode*, std::map<std::string, PropertyValue, std::less<>>> mode_user_props_;
+  std::map<std::string, PropertyValue, std::less<>> mode_user_prop_defaults_;
```
jjcherry56, line 306: All these default values maps will get messy very quickly.

search/Property.cc
```diff
@@ -1360,6 +1372,115 @@ Properties::defineProperty(std::string_view property,
...
+{
+  std::string name(property);
+  mode_user_prop_defaults_[name] = userPropertyDefault(type);
+  defineProperty(property,
+                 [this, name] (const Mode *mode, Sta *) -> PropertyValue {
+    auto oi = mode_user_props_.find(mode);
+    if (oi != mode_user_props_.end()) {
+      auto pi = oi->second.find(name);
+      if (pi != oi->second.end())
+        return pi->second;
```
jjcherry56: pretty yucky chain of iterators.
oi? pi? where on earth do those names come from? ai slop?

search/Property.i
```diff
@@ -137,6 +137,40 @@ mode_property(Mode *mode,
   return properties.getProperty(mode, property);
 }
 
+void
```
jjcherry56, line 141: All of this code should be generic of object type.

include/sta/Property.hh
dsengupta0628: Okay now using a class class UserPropertyKey  and one single member in Properties class  std::map<UserPropertyKey, PropertyValue> user_prop_values_ for any generic object to address this. 

Only things: I am now using template<class TYPE> void defineUserProperty based on the above suggestion- The other defineProperty(std::string_view property,..) functions (not user property), are object specific not templated- but I am not touching them now. I guess I was trying to match that style.

search/Property.cc
dsengupta0628: sorry- yes these were short for object iterator, property iterator and default iterator. Terrible short forms- fixing!

search/Property.i
dsengupta0628: For now I have made one generic define/set_user_property_cmd (in search/Property.i) instead of define/set_scene_user_property or define/set_mode_user_property;  But within this generic function, still only "scene" and "mode" branches exist; anything else would show error. 

To be able to make this command truly generic for all object types, I need explicit instantiation line for them in Property.cc and call them appropriately from define/set_user_property_cmd. That is likely out of scope of this work. but if you still want that, I can make the change.

include/sta/Property.hh
```diff
@@ -124,6 +238,24 @@ public:
...
   void defineProperty(std::string_view property,
                       const PropertyRegistry<const Clock *>::PropertyHandler &handler);
+  void defineProperty(std::string_view property,
+                      const PropertyRegistry<const Scene *>::PropertyHandler &handler);
+  void defineProperty(std::string_view property,
+                      const PropertyRegistry<const Mode *>::PropertyHandler &handler);
+
+  // User-defined, per-object mutable properties. defineUserProperty registers
+  // a property of the given value type ("bool", "float" or "string") with a
+  // typed default; setUserProperty overrides the value on one object. The
```
parallaxsw: again, what is the point of having a default value?

include/sta/Property.hh
```diff
@@ -124,6 +238,24 @@ public:
...
+  // property is read through the same registry path as every other property so
+  // get_property / get_* -filter work unchanged.
+  template<class TYPE>
+  void defineUserProperty(std::string_view object_type,
```
parallaxsw: I am not convinced there is any point in  defineProperty? does user really need to be in all of the function/command names?

search/Property.cc
```diff
@@ -1360,6 +1373,114 @@ Properties::defineProperty(std::string_view property,
...
+PropertyValue
+Properties::coerceUserValue(PropertyValue::Type type,
+                            std::string_view value)
+{
+  switch (type) {
+  case PropertyValue::Type::bool_:
+    return PropertyValue(value == "true" || value == "1");
+  case PropertyValue::Type::float_: {
+    auto [number, valid] = stringFloat(std::string(value));
+    return PropertyValue(valid ? number : 0.0f, sta_->units()->scalarUnit());
```
parallaxsw: what is checking for valid and reporting warnings or errors?

search/Property.i
```diff
@@ -137,6 +138,39 @@ mode_property(Mode *mode,
...
+define_user_property_cmd(const char *object_type,
```
parallaxsw: define_property

include/sta/Property.hh
dsengupta0628: Suppose user defined a bool property "active" on scene object, but only set the value for one or two particular scenes among many scenes defined in the design. What would the rest of the scenes get the value for that property? I removed that ambiguity by defining the default values based on the property type- hence the default values. 

The default is what getProperty returns for objects the property was never set on - the filter evaluates every object in the set, so unset objects need some value. Making it typed rather than empty is what lets -filter {active ==  false} match unset scenes and get_property return 0/false for them.

search/Property.i
dsengupta0628: Will update to define_property and set_property instead of *_user_* versions

search/Property.cc
dsengupta0628: Fixed now

include/sta/Property.hh
dsengupta0628: Agreed on the names -I'll drop "user" throughout: tcl define_property / set_property, and the C++ side becomes overloads of the existing defineProperty alongside a new setProperty.

On removing the define step: I considered dropping define and letting set_property implicitly create a property on first use (all values as strings), but that loses these 2 benefits the define provides:

1. The filter loop calls getProperty on every object in the set with no try/catch (filterObjects, FilterObjects.cc), so objects that never had the property set must still return a value or the whole command errors out - and filtering a partially-set  property is the main use case: get_scenes -filter {active == true} where only scene1 was ever set has to evaluate active on scene2 too. Some fallback value for unset objects is therefore mandatory; define makes it an explicit typed default (unset bool reads as false) instead of an invented empty string.
2. `set_property scene1 activ true` now errors immediately ("property not defined") instead of silently minting a new property on a typo (that would happen without the define step)

include/sta/Property.hh
dsengupta0628: Dropped now. Defaults are problematic as it doesn't consider that properties maybe deliberately unset. If user wants, they need to explicitly set property value to all instances of the object class

include/sta/Property.hh
parallaxsw: I agree with leaving define_property. My intent was to question the use of "user" in the commands and you have removed it.

search/Property.cc
```diff
@@ -1360,6 +1365,123 @@ Properties::defineProperty(std::string_view property,
...
+    return PropertyValue(std::string());
+  else
+    sta_->report()->error(2210, "unknown property type '{}' (use bool, float or string).",
+                          type);
+  return PropertyValue();
+}
+
+PropertyValue
+Properties::coercePropertyValue(PropertyValue::Type type,
+                            std::string_view value)
```
parallaxsw: indent

search/Property.cc
```diff
@@ -1360,6 +1365,123 @@ Properties::defineProperty(std::string_view property,
...
+      sta_->report()->error(2215, "'{}' is not a float property value.", value);
+    return PropertyValue(number, sta_->units()->scalarUnit());
+  }
+  default:
+    return PropertyValue(std::string(value));
+  }
+}
+
+PropertyKey::PropertyKey(const void *object,
+                                 std::string_view object_type,
```
parallaxsw: indent

search/Property.cc
```diff
@@ -1360,6 +1365,123 @@ Properties::defineProperty(std::string_view property,
...
+    < std::tie(key.object_, key.object_type_, key.property_);
+}
+
+template<class TYPE>
+void
+Properties::defineProperty(std::string_view object_type,
+                               std::string_view property,
+                               std::string_view value_type)
+{
+  prop_values_[PropertyKey(nullptr, object_type, property)] =
```
parallaxsw: I don't understand the point of storing a property value for a null object or the propertyTypeValue function.
If the get_property object is null it should not return a value.

dsengupta0628: Hmm- will remove. However, I will need one double map as shown below: (giving disclaimer as you don't  prefer those usually)

prop_types_ will be a std::map<std::pair<std::string,std::string>, PropertyValue::Type>  with  key as (object_type, property) pair, value = property type enum. 

Example 
If we have
define_property -object_type scene -type bool active
          --> means it is stored as
  prop_types_[{"scene", "active"}] = PropertyValue::Type::bool_

this way PropertyKey class is a bit simplified; it need not store the object type as  a string- the object's pointer is enough info.

So my LHS can become simpler like this:  prop_values_[PropertyKey(object, property)] instead of 3 members in PropertyKey

(I think this was remnant from my default implementation)

search/Property.i
```diff
@@ -137,6 +138,39 @@ mode_property(Mode *mode,
   return properties.getProperty(mode, property);
 }
 
+void
+define_property_cmd(const char *object_type,
+                         const char *property,
```
parallaxsw: indent

search/Property.i
```diff
@@ -137,6 +138,39 @@ mode_property(Mode *mode,
...
+  else if (object_type_view == "mode")
+    properties.defineProperty<Mode>(object_type, property, type);
+  else
+    Sta::sta()->report()->error(2209, "define_property -object_type {} not supported.",
+                                object_type);
+}
+
+void
+set_property_cmd(void *object,
+                      const char *object_type,
```
parallaxsw: indent

PR #461, opened 2026-07-11
SMALLFIX: Honor user activity map in `Power::findActivity`

akashlevy:
## Summary
- Before returning unknown/zero activity, `Power::findActivity` now checks `user_activity_map_` (via `hasUserActivity` / `userActivity`).
- Pins annotated with `set_power_activity` are reported correctly even when activity cannot be propagated (e.g. flop or macro outputs).

## Motivation
User-annotated activities were ignored for pins that never appear in the propagated `activity_map_`, so reports showed 0 / unknown despite explicit annotations.

## Notes
- No dedicated regression yet (happy to add one if desired).

## Test plan
- [x] Manually: `set_power_activity` on a flop/macro output, then `report_activity_annotation` / power report shows the user values
- [x] Existing power / VCD regressions still pass

### DIFF
--- a/power/Power.cc
+++ b/power/Power.cc
@@ -1604,6 +1604,8 @@ Power::findActivity(const Pin *pin)
     if (activity && activity->origin() != PwrActivityOrigin::unknown)
       return *activity;
   }
+  if (hasUserActivity(pin))
+    return userActivity(pin);
   return PwrActivity(0.0, 0.0, PwrActivityOrigin::unknown);
 }
 
--- a/test/find_activity_user.ok
+++ b/test/find_activity_user.ok
@@ -1,0 +1,8 @@
+input           1
+unannotated    19
+Annotated pins:
+ user r1/Q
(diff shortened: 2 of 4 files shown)


parallaxsw:
The test passes without the change

akashlevy:
You are right, the test is bad. However, the fix should still be applied. Can you reopen this so I can make a proper test case?

parallaxsw:
I think you can re-open it.

I recall fixing this problem some time ago but I can't find the commit.
I have a regression for a macro with a user activity on an output with no timing arcs
to it. get_property activity reports the correct value (it is a root so the activity appears in activity_map_.

Why don't you file an issue with a test case that actually shows a problem?

akashlevy:
I'm not able to re-open this PR, the settings seem to prevent that.

But yes, I can file an issue with a test case first and then you can apply the fix if you think it's apt.

IIRC, it was a corner case we hit in one of our power scripts, which affected some of our numbers adversely

PR #465, opened 2026-07-11
FEATURE: Add `set_disable_clock_gating_check` for cells, instances, and pins

akashlevy:
## Summary
- Add `set_disable_clock_gating_check` / `unset_disable_clock_gating_check` accepting Liberty cells, instances, ports, and pins.
- Keep `set_disable_inferred_clock_gating` as a compatibility alias.
- Persist disables through `write_sdc`.
- Add `disable_clock_gating_check` regression.

## Motivation
Inferred clock-gating check disables previously only covered instances/pins and were not named consistently with SDC-style `set_disable_clock_gating_check`. Cell-level disable is needed when all instances of a clock-gating cell should skip the check.

## Notes
- Docs still TODO if you want a man page / README mention.

## Test plan
- [x] `./regression disable_clock_gating_check` (local)


akashlevy:
Will add docs at the end.

search/Search.tcl
```diff
@@ -978,42 +978,78 @@ proc_redirect report_clock_min_period {
...
+    unset_disable_clock_gating_check_pin [get_port_pin $port]
+  }
   foreach pin $pins {
     unset_disable_clock_gating_check_pin $pin
   }
 }
 
 ################################################################
 
+define_cmd_args "set_disable_inferred_clock_gating" { objects }
```
parallaxsw: this should report a deprecated warning and have a comment stating the date it was deprecated (for removal at a later date). grep deprecated in the tcl files for examples

test/disable_clock_gating_check.tcl
```diff
@@ -0,0 +1,47 @@
...
+puts "-- back-compat alias --"
+set_disable_inferred_clock_gating [get_cells cg]
+unset_disable_inferred_clock_gating [get_cells cg]
+puts "ok"
+
+puts "-- write_sdc --"
+set_disable_clock_gating_check [list [get_lib_cells AND2] [get_cells cg] [get_pins cg/B]]
+write_sdc results/disable_clock_gating_check.sdc
+set fp [open results/disable_clock_gating_check.sdc r]
+foreach line [split [read $fp] "\n"] {
```
parallaxsw: use test/helpers.tcl report_file. see examples in existing regressions

test/disable_clock_gating_check.tcl
```diff
@@ -0,0 +1,47 @@
+read_liberty disable_clock_gating_check.lib
+read_verilog disable_clock_gating_check.v
+link_design top
+create_clock -name clk -period 1.0 [get_ports clk]
+set_input_delay -clock clk 0 [get_ports {en d}]
+
+puts "-- libcell --"
```
parallaxsw: all this does is check the arg parsing. it does not check the functionality.
i think they should be separate tests for each use and check they actually
do something (I have private tests for the previous command but not these
obviously).

akashlevy:
@jjcherry56 All your comments should be addressed. Let me know if anything else is needed.

include/sta/Sdc.hh
```diff
@@ -708,10 +708,19 @@ public:
...
   void disableClockGatingCheck(Instance *inst);
   void disableClockGatingCheck(Pin *pin);
+  void disableClockGatingCheck(LibertyCell *cell);
   void removeDisableClockGatingCheck(Instance *inst);
   void removeDisableClockGatingCheck(Pin *pin);
+  void removeDisableClockGatingCheck(LibertyCell *cell);
   bool isDisableClockGatingCheck(const Pin *pin) const;
   bool isDisableClockGatingCheck(const Instance *inst) const;
+  bool isDisableClockGatingCheck(const LibertyCell *cell) const;
+  const InstanceSet *disabledClockGatingChecksInst() const
```
parallaxsw: why return pointers? why not refs?
also, if the function is longer than a line it shouldn't be in the header

akashlevy: Moved these to Sdc.cc and switched them to refs.

akashlevy:
@jjcherry56 accessors now return refs and live in Sdc.cc.

PR #469, opened 2026-07-11
BIGFEATURE: Add `set_path_margin` command

akashlevy:
## Summary
- Implements `set_path_margin` path exception and applies the margin in path-end slack reporting.
- Writes the exception via `write_sdc` and includes a regression.

## Test plan
- [x] `./regression set_path_margin`
- [x] CI green on draft


akashlevy:
TODO: docs

test/set_path_margin.tcl
```diff
@@ -0,0 +1,90 @@
+# set_path_margin: per-path slack adjustment on the capture clock.
```
parallaxsw: break this up into separate tests named set_path_margiin1,2 etc.

search/ReportPath.cc
```diff
@@ -1134,6 +1134,10 @@ ReportPath::reportJson(const PathEnd *end,
                  delayAsFloat(end->checkCrpr(this)));
     result += sta::format("  \"margin\": {:.3e},\n",
                  delayAsFloat(end->margin(this)));
+    float path_margin = end->targetClkPathMargin(this);
+    if (path_margin != 0.0)
```
parallaxsw: I think you should keep the path margin exception in the path end, define PathEnd::hasPathMargin and use that to decide when to priint instead of non-zero.

test/set_path_margin.tcl
akashlevy: Done — split into `set_path_margin1`–`set_path_margin6`, each self-contained so leftover exceptions from earlier sections no longer leak into later checks.

search/ReportPath.cc
akashlevy: Done — `PathMargin*` is now stored on `PathEndClkConstrained` (passed from `VisitPathEnds`), with `PathEnd::hasPathMargin()` / `pathMargin()`. Text and JSON reporting gate on `hasPathMargin()` instead of a non-zero float.

akashlevy:
Addressed review feedback and rebased onto current `master` (conflicts resolved):

- Store `PathMargin` on the path end; use `PathEnd::hasPathMargin()` for reporting
- Split regression into `set_path_margin1`–`6`
- Added `ChangeLog` entry for `set_path_margin`

akashlevy:
Added `set_path_margin` docs to `OpenSTA.fodt` / `OpenSTA.pdf`, and updated `unset_path_exceptions` to mention path margins.

akashlevy:
@jjcherry56 ready for review

PR #470, opened 2026-07-11
SMALLFIX: Free SWIG-allocated Seq containers

akashlevy:
## Summary
- SWIG `%typemap(freearg)` owns borrow-only `Seq*` inputs (`PinSeq`, `InstanceSeq`, `CellSeq`, …). Callees no longer `delete` those containers.
- `%exception` now uses `SWIG_fail` instead of `return TCL_ERROR`, so `freearg` still runs when `FilterExpr` throws. That was the `get_filter` leak: `tclListSeqPtr` allocated an `InstanceSeq`, the invalid `name ~= …` path threw, and the wrapper returned without deleting it.
- Do **not** `freearg` types C++ keeps (`ExceptionThruSeq`, `PinSet`, `ClockSet`, `InstanceSet`, `NetSet`). `make_false_path` stores the thru seq; freeing it would use-after-free.

The double-free was the other side of the same bug: `freearg` plus the old C++ `delete` on a successful `get_* -filter`. Master never had `freearg`, so that double-free only showed up once the typemaps were added.

## Test plan
- [x] `./regression swig_seq_double_free swig_seq_leak get_filter`
- [x] `./regression all` (59 public tests)


parallaxsw:
I agree this is a cleaner pattern, but I do not see anywhere there is a "double free".

akashlevy:
We had it show double frees in our flows, but I can't recall how it was triggered. I will see if I can create a test case.

parallaxsw:
This causes a leak in the get_filter regression
==1074== 64 bytes in 1 blocks are indirectly lost in loss record 10 of 116
==1074==    at 0x48859FC: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-arm64-linux.so)
==1074==    by 0x2427AF: std::__new_allocator<sta::Instance const*>::allocate(unsigned long, void const*) (new_allocator.h:151)
==1074==    by 0x23C473: allocate (allocator.h:198)
==1074==    by 0x23C473: allocate (alloc_traits.h:482)
==1074==    by 0x23C473: std::_Vector_base<sta::Instance const*, std::allocator<sta::Instance const*> >::_M_allocate(unsigned long) (stl_vector.h:381)
==1074==    by 0x2366DF: void std::vector<sta::Instance const*, std::allocator<sta::Instance const*> >::_M_realloc_insert<sta::Instance const*>(__gnu_cxx::__normal_iterator<sta::Instance const**, std::vector<sta::Instance const*, std::allocator<sta::Instance const*> > >, sta::Instance const*&&) (vector.tcc:459)
==1074==    by 0x23199F: sta::Instance const*& std::vector<sta::Instance const*, std::allocator<sta::Instance const*> >::emplace_back<sta::Instance const*>(sta::Instance const*&&) (vector.tcc:123)
==1074==    by 0x22BCFB: std::vector<sta::Instance const*, std::allocator<sta::Instance const*> >::push_back(sta::Instance const*&&) (stl_vector.h:1299)
==1074==    by 0x2241BF: std::vector<sta::Instance const*, std::allocator<sta::Instance const*> >* sta::tclListSeqPtr<sta::Instance const*>(Tcl_Obj*, swig_type_info*, Tcl_Interp*) (StaAppTCL_wrap.cxx:1918)
==1074==    by 0x1DBDB3: _wrap_filter_insts (StaAppTCL_wrap.cxx:31444)
==1074==    by 0x48FCB0F: TclNRRunCallbacks (in /usr/lib/aarch64-linux-gnu/libtcl8.6.so)
==1074==    by 0x48FDE07: ??? (in /usr/lib/aarch64-linux-gnu/libtcl8.6.so)
==1074==    by 0x48FEA8F: Tcl_Eval (in /usr/lib/aarch64-linux-gnu/libtcl8.6.so)
==1074==    by 0x2443AF: sta::sourceTclFile(char const*, bool, bool, Tcl_Interp*) (StaMain

akashlevy:
The valgrind `get_filter` leak was real. `tclListSeqPtr` allocates the `InstanceSeq`, then `get_cells -filter {name ~= *r1*}` throws. `%exception` used `return TCL_ERROR`, which skips SWIG's `fail:` cleanup, so `freearg` never ran.

That is also why FilterObjects used to `delete` the seq *before* parsing: so a thrown parse error would not leak. Removing that delete without making `%exception` go through `SWIG_fail` leaked on the error path. Keeping both the C++ `delete` and `freearg` double-frees on the success path (`get_cells -filter liberty_cell==…`).

This update:
- uses `SWIG_fail` in `Exception.i` so `freearg` runs after STA exceptions
- leaves `freearg` only on borrow-only `Seq*` inputs (not `ExceptionThruSeq` / `PinSet` / etc. that C++ stores)
- adds `swig_seq_leak` (invalid `~=` on each filter wrapper, including `sta::filter_insts`) and `swig_seq_double_free` (successful filter/fanin/`write_verilog`/`report_path_ends`/`set_false_path`)

akashlevy:
@jjcherry56 I now have an example of the double free issue. The updated fix properly manages all the SWIG memory, without any leaks or double-frees. Check it out when you have time

PR #473, opened 2026-07-11
feat(ci): expand CI with macOS and CentOS 7 (Docker) platform coverage

akashlevy:
## Summary
Merges the useful parts of Silimate `.github/workflows/tests.yaml` into upstream `ci.yml` (instead of adding a second overlapping workflow).

**What this keeps**
- Existing native Ubuntu job (CUDD tarball, Ninja build, install artifacts, full `./regression`), renamed `linux-ubuntu`.
- Install artifact upload with upstream `retention-days: 1`, named `artifact-ubuntu`.

**What this adds**
- `linux-centos`: CentOS 7 via `docker build -f Dockerfile.centos7` (glibc/legacy coverage).
- `macos`: Brewfile-based native macOS build/test.

**What this intentionally drops from Silimate `tests.yaml`**
- Separate `tests.yaml` workflow (would double-run CI on every PR).
- Docker Ubuntu 22.04 / 24.04 jobs (redundant with native Ubuntu).
- `./regression -collections` fallback (collections is not part of upstream).
- Older `actions/checkout@v4` (kept upstream `@v7`).
- Separate `brew trust` CI step.
- Manual Dockerfile copy / `docker rm` / ignored `docker cp` return codes.

**Brew trust**
- Brewfile declares `brew "mht208/formal/cudd", trusted: true` so `brew bundle` applies formula-level trust. No extra CI step.

## Test plan
- [x] Confirm CI jobs: `linux-ubuntu`, `linux-centos`, `macos`
- [x] Confirm artifacts: `artifact-ubuntu`, `result-ubuntu`, `result-centos7`, `result-macos`
- [x] Confirm macOS installs CUDD via `brew bundle` alone

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
   pull_request:
 
 jobs:
+  # Primary native Linux build used for install artifacts.
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -50,6 +51,84 @@ jobs:
         uses: actions/upload-artifact@v7
         if: ${{ !cancelled() }}
         with:
-          name: result
+          name: result-ubuntu-latest
           path: |
             test/results/*
+          retention-days: 1
```
kazutoiris: There is no need to manually configure `retention-days`, especially this result requires the same retention period as the logs.

.github/workflows/ci.yml
```diff
@@ -50,6 +51,84 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          submodules: true
+
+      - name: Build image
+        run: |
+          cp Dockerfile.centos7 Dockerfile
+          docker build --tag sta-centos7 .
```
kazutoiris: Use `docker build -f`, instead of manually copying files.

.github/workflows/ci.yml
```diff
@@ -50,6 +51,84 @@ jobs:
...
+
+      - name: Test in container
+        run: |
+          set +e
+          docker run --name sta-test --entrypoint /bin/bash sta-centos7 \
+            -c "cd /OpenSTA/test && ./regression || (cat results/diffs && exit 1)"
+          status=$?
+          mkdir -p test/results
+          docker cp sta-test:/OpenSTA/test/results/. test/results/ || true
+          docker rm -f sta-test || true
```
kazutoiris: No need to remove the container manually. Also, do not ignore the return code.

.github/workflows/ci.yml
```diff
@@ -50,6 +51,84 @@ jobs:
...
+          docker build --tag sta-centos7 .
+
+      - name: Test in container
+        run: |
+          set +e
+          docker run --name sta-test --entrypoint /bin/bash sta-centos7 \
+            -c "cd /OpenSTA/test && ./regression || (cat results/diffs && exit 1)"
+          status=$?
+          mkdir -p test/results
+          docker cp sta-test:/OpenSTA/test/results/. test/results/ || true
```
kazutoiris: Do not ignore the return code.

kazutoiris (changes requested):
I've provided some suggestions for your reference to implement the fixes.

BTW, the job names can be modified. Previously, with only a single job, naming it `build` was perfectly fine. However, now that there are multiple jobs, I recommend naming them as follows:

- `linux-ubuntu`
- `linux-centos`
- `macos`

akashlevy:
Addressed review feedback:

- Renamed jobs to `linux-ubuntu`, `linux-centos`, and `macos`
- Removed `retention-days` from test-result uploads (left install artifact as upstream)
- Switched to `docker build -f Dockerfile.centos7`
- Replaced named-container `docker cp`/`docker rm` with a volume mount and `--rm`, so regression return codes are not ignored

.github/workflows/ci.yml
```diff
@@ -50,6 +51,76 @@ jobs:
         uses: actions/upload-artifact@v7
         if: ${{ !cancelled() }}
         with:
-          name: result
+          name: result-ubuntu-latest
```
kazutoiris: `-latest` is redundant.

.github/workflows/ci.yml
kazutoiris: You also need to modify the name here to avoid ambiguity.

kazutoiris (changes requested):
I've provided some suggestions for your reference to implement the fixes.

akashlevy:
Addressed the latest review notes:

- Renamed install artifact `artifact` → `artifact-ubuntu`
- Renamed test artifact `result-ubuntu-latest` → `result-ubuntu`

kazutoiris (approved):
LGTM.

akashlevy:
@kazutoiris thanks for reviewing. Do you (or @jjcherry56) think it makes sense to upload artifacts for the non-Ubuntu platforms or no?

kazutoiris:
> @kazutoiris thanks for reviewing. Do you (or @jjcherry56) think it makes sense to upload artifacts for the non-Ubuntu platforms or no?

There's no need to upload other platforms artifacts. This CI is used only for validation (not for releases), so retaining a single artifact is sufficient for troubleshooting.

akashlevy:
Sounds good, thanks!

PR #475, opened 2026-07-16
write_path_spice: match side input values to the path arc's transitions

### LINKED ISSUE #474: write_path_spice ties non-unate side inputs (xor/xnor, mux selects) without honoring the when-condition of the arc the path used.
I found a bug, and I have a fix, and I have created a standalone test and verification.
The patches that I use for my eda tree are here: https://github.com/bpdegnan/macportseda/tree/main/cad/OpenSTA/files
I have a stand-alone example of the bug using SKY130.  It was found during verification of a RISC-V where we noticed a ~10% descrepency  
[opensta.tgz](https://github.com/user-attachments/files/30077840/opensta.tgz)

Example, the forced B:fall -> X:rise path, the liberty's when A requirement, and the two damning lines (v1 x0/A 0 0.000 in the deck vs the v/^ edges in the report). 
The precise issue is that first Cudd_FirstCube of the Boolean difference in gatePortValues(), ignores transition directions, unknown falls through to tie-low. 
With sta/ngspice/awk, you can run "run_repro.sh" and see the validated fix.  PR to follow

bpdegnan:
Fixes #474.

`gatePortValues()` chooses side-input tie values from the first CUDD cube of the Boolean difference `d(f)/d(input)`. That cube sensitizes the gate combinationally but ignores the transition directions of the arc the path actually used, with two visible failure shapes:

- **xor2/xnor2**: the Boolean difference is a tautology, so every side variable comes back don't-care (`LogicValue::unknown`), which falls through to the tie-low case in `writeSubcktInstVoltSrcs()`. Wrong whenever the reported path used the when-condition that requires the side high.
- **mux select arcs**: `d(f)/d(sel)` is `(data_a XOR data_b)`, and the first cube is an arbitrary data assignment — inverted relative to the reported output edge about half the time.

From a mis-tied gate onward the simulated chain switches in the opposite direction from the arcs STA summed: edge-qualified arrival measurements fail, and the deck effectively sums delays from the wrong rise/fall tables. Issue #474 has a self-contained single-gate SKY130 reproduction (sta + ngspice + awk, ~5 s) whose verdict script distinguishes buggy from fixed behavior.

**The fix** constrains the side-input condition to the cofactor pair matching the arc's polarity — `f1 & !f0` when the gate input and driver edges agree (non-inverting), `f0 & !f1` when they differ (inverting) — threading the gate input `RiseFall` from the path stage (already available via `stageGateInputPath`) into `gatePortValues()`. It also releases the CUDD nodes the previous code leaked (`Cudd_Ref` on the Boolean difference after `Cudd_GenFree`, never deref'd).

**Validation**
- The #474 repro flips from `VERDICT: BUG` to `VERDICT: FIXED`: the xor side input is tied to VPWR, and the edge-qualified `t_arrive` measure succeeds and equals the direction-agnostic `t_arrivx`.
- On two synthesized RISC-V-class cores where we found the problem (write_path_spice decks correlated against ngspice), regenerating 30 path decks covering 57 non-unate path gates (xor2/xnor2 side inputs and mux2 select arcs, previously 26 decks mis-tied): every tie now matches the reported arc, and per-path STA-vs-SPICE error on the worst-affected core went from about -10% to the same few-percent band as unaffected cores.

The test suite has no write_path_spice coverage today; happy to add a regression test (netlist + golden deck diff, or the #474 repro reduced to your preferred harness) if you'd like one here.


jjcherry56:
You need to add the test case from the issue to the PR in /test.
See doc/CodingGuidlines.txt

bpdegnan:
> You need to add the test case from the issue to the PR in /test. See doc/CodingGuidlines.txt

the test is pushed and passes through test/regression for xor2.

PR #476, opened 2026-07-16
PrimaDelayCalc: handle degenerate parasitic networks (fixes STA-1752 "G matrix is singular")

### LINKED ISSUE #4355: GRT/STA hit "G matrix is singular" for two asap7 designs
### Subject

[Stage]: Global Router.

### Describe the bug

When running the RC correlation flow on ASAP7 with MAX_ROUTING_LAYER=M9, two designs (gcd-ccs and swerv-wrapper) run into the following STA error in GRT:

```
[ERROR STA-1752] G matrix is singular.
```

In the case of swerv-wrapper, you get a stack:

```
[ERROR STA-1752] G matrix is singular.
terminate called after throwing an instance of 'std::runtime_error'
  what():  STA-1752
Signal 6 received
Stack trace:
 0# 0x0000000000EFA697 in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x0000000000042520 in /lib/x86_64-linux-gnu/libc.so.6
 2# pthread_kill in /lib/x86_64-linux-gnu/libc.so.6
 3# raise in /lib/x86_64-linux-gnu/libc.so.6
 4# abort in /lib/x86_64-linux-gnu/libc.so.6
 5# 0x00000000000A2B9E in /lib/x86_64-linux-gnu/libstdc++.so.6
 6# 0x00000000000AE20C in /lib/x86_64-linux-gnu/libstdc++.so.6
 7# 0x00000000000AE277 in /lib/x86_64-linux-gnu/libstdc++.so.6
 8# 0x00000000000AE4D8 in /lib/x86_64-linux-gnu/libstdc++.so.6
 9# 0x000000000096C30A in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
10# sta::dbStaReport::errorMsg(int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
11# 0x00000000010D6BBB in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
12# sta::PrimaDelayCalc::primaReduce() in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
13# sta::PrimaDelayCalc::simulate() in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
14# sta::PrimaDelayCalc::gateDelays(std::vector<sta::ArcDcalcArg, std::allocator<sta::ArcDcalcArg> >&, std::map<sta::Pin const*, unsigned long, sta::PinIdLess, std::allocator<std::pair<sta::Pin const* const, unsigned long> > > const&, sta::Scene const*, sta::MinMax const*) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
15# sta::PrimaDelayCalc::gateDelay(sta::Pin const*, sta::TimingArc const*, sta::Delay const&, float, sta::Parasitic const*, std::map<sta::Pin const*, unsigned long, sta::PinIdLess, std::allocator<std::pair<sta::Pin const* const, unsigned long> > > const&, sta::Scene const*, sta::MinMax const*) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
16# sta::GraphDelayCalc::findDriverArcDelays(sta::Vertex*, sta::MultiDrvrNet const*, sta::Edge*, sta::TimingArc const*, sta::Scene const*, sta::MinMax const*, sta::ArcDelayCalc*, std::map<sta::Pin const*, unsigned long, sta::PinIdLess, std::allocator<std::pair<sta::Pin const* const, unsigned long> > >&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
17# sta::GraphDelayCalc::findDriverEdgeDelays(sta::Vertex*, sta::MultiDrvrNet const*, sta::Edge*, sta::ArcDelayCalc*, std::map<sta::Pin const*, unsigned long, sta::PinIdLess, std::allocator<std::pair<sta::Pin const* const, unsigned long> > >&, std::array<bool, 2ul>&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
18# sta::GraphDelayCalc::findDriverDelays1(sta::Vertex*, sta::MultiDrvrNet*, sta::ArcDelayCalc*, std::map<sta::Pin const*, unsigned long, sta::PinIdLess, std::allocator<std::pair<sta::Pin const* const, unsigned long> > >&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
19# sta::GraphDelayCalc::findDriverDelays(sta::Vertex*, sta::ArcDelayCalc*, std::map<sta::Pin const*, unsigned long, sta::PinIdLess, std::allocator<std::pair<sta::Pin const* const, unsigned long> > >&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
20# sta::GraphDelayCalc::findVertexDelay(sta::Vertex*, sta::ArcDelayCalc*, bool) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
21# 0x00000000011D03E4 in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
22# sta::DispatchQueue::dispatch_thread_handler(unsigned long) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
23# 0x00000000000DC253 in /lib/x86_64-linux-gnu/libstdc++.so.6
24# 0x0000000000094AC3 in /lib/x86_64-linux-gnu/libc.so.6
25# clone in /lib/x86_64-linux-gnu/libc.so.6
Elapsed time: 1:23.68[h:]min:sec. CPU time: user 1083.66 sys 10.34 (1307%). Peak memory: 3967536KB.
```

The default runs with MAX_ROUTING_LAYER=M7 don't encounter the errors.

### Expected Behavior

Successful GRT run

### Environment

```shell
orfs git commit tag: afad87da8d1ebd6ef1bc94288b4204420975a549
```

### To Reproduce

gcd-ccs tarball: https://drive.google.com/file/d/1wlIEAq_B1YAkglm5aoEqLNkg2DUus21E/view?usp=drive_link
swerv_wrapper tarball: https://drive.google.com/file/d/1Tg5pAKO_2JtV8BfP0E4IH39YGhaKP2Uk/view?usp=drive_link

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

dsengupta0628:
## Summary

Fixes `[ERROR STA-1752] G matrix is singular` — and the `SIGABRT` it caused on multithreaded runs — when the Prima/CCS delay calculator is handed degenerate parasitic networks from resistance-aware global route.

Fixes The-OpenROAD-Project/OpenROAD-flow-scripts#4355.

## Root cause

Prima builds a conductance matrix `G` (which requires `1 / R`) and solves it. Two malformed inputs make `G` singular, for different reasons:

- **Cause A — floating node:** a node with a capacitor but no resistor tying it to the driver. Since nodes are connected only through resistors, it has no resistive path to the driver → an empty equation row.
- **Cause B — zero-ohm short:** a resistor with `R = 0`, so `conductance = 1 / 0 = ∞` corrupts the matrix. The two endpoints are really one node.

image

In above figure, two points are tied together only through a resistor; a capacitor just leaks charge to ground. Left: a healthy net every solver can handle. Middle and right: the two malformed inputs that caused STA-1752

The crash-vs-clean-error split was purely threading: on small designs the failing calc runs on the main thread and the error surfaces as a normal Tcl error; on large designs it runs on a worker thread, where the raised error escapes the thread → `std::terminate` → `SIGABRT`.

> In these ORFS runs, Cause B traces to the asap7 `setRC.tcl` defining RC for M1–M7 only while GRT routed up to M9, so M8/M9 segments defaulted to `R = 0`. That's an upstream data gap; this PR makes Prima robust to such input regardless of source.

## Fix

A single walk from the driver over resistors, in `dcalc/PrimaDelayCalc.cc`:

- Nodes never reached (no resistive path to the driver) are never added to the system → resolves **Cause A**.
- When the walk crosses a zero/negative resistor, the far node is given the **same slot** as the near node → merges shorted nodes for **Cause B**; the `1 / R` term is skipped for the short.

After the walk, every remaining node has a resistive path to the driver, so `G` is always solvable and the computed delays are correct. The pass is a no-op on well-formed nets.


test/get_scenes.ok
```diff
@@ -30,4 +30,4 @@ scene2
 [get_property scene2 active] (set false)
```
dsengupta0628, line 30: Not sure how this file got added. In my fork I see it exactly same as in upstream.

dcalc/PrimaDelayCalc.cc
```diff
@@ -569,14 +604,16 @@ PrimaDelayCalc::stampEqns()
...
-      float resistance = parasitics_->value(resistor);
-      stampConductance(node_idx1, node_idx2, 1.0 / resistance);
+    auto itr1 = node_index_map_.find(parasitics_->node1(resistor));
+    auto itr2 = node_index_map_.find(parasitics_->node2(resistor));
+    float resistance = parasitics_->value(resistor);
+    // Skip a resistor with an excluded endpoint, endpoints on the same system
+    // index (self loop or merged short), or a non-positive (short) resistance.
+    if (itr1 != node_index_map_.end()
+        && itr2 != node_index_map_.end()
+        && itr1->second != itr2->second
```
parallaxsw: not a fan of seeing second, especially when repeated.
either bind node vars or use findKeyValue so this isn't so c++ iterator ugly

dcalc/PrimaDelayCalc.cc
```diff
@@ -468,35 +468,70 @@ PrimaDelayCalc::findNodeCount()
...
-      size_t node_idx = node_index_map_.size();
-      node_index_map_[node] = node_idx;
-      const Pin *pin = parasitics_->pin(node);
-      if (pin) {
-        pin_node_map_[pin] = node_idx;
-        debugPrint(debug_, "ccs_dcalc", 1, "pin {} node {}",
-                   network_->pathName(pin), node_idx);
+  // A parasitic node enters the conductance matrix G only if it is reachable
+  // from a driver through resistors. Nodes couple only through resistors
+  // (capacitors are all stamped to ground), so a node with no resistive path to
```
parallaxsw: No, nodes are not all stamped to ground when the signal integrity branch supports coupling capacitors. So delete that part of the comment.

dcalc/PrimaDelayCalc.cc
```diff
@@ -468,35 +468,70 @@ PrimaDelayCalc::findNodeCount()
...
+  // a driver would leave an all-zero row in G and make it singular; such a node
+  // carries no current and does not affect timing, so it is left out. A zero
+  // (or negative) resistance resistor is an ideal short, so its two nodes share
+  // one index (stamping 1/resistance would be infinite); the short itself is
+  // skipped when stamping G, see stampEqns().
+  ParasiticNodeResistorMap resistor_map =
+      parasitics_->parasiticNodeResistorMap(parasitic_network_);
+  std::vector<ParasiticNode *> queue;
+  auto place_node = [&](ParasiticNode *node, size_t index) {
```
parallaxsw: I don't see the point of defining a lambda when there is only one caller.
But this function could use some refactoring

dcalc/PrimaDelayCalc.cc
```diff
@@ -468,35 +468,70 @@ PrimaDelayCalc::findNodeCount()
...
+          place_node(next_node,
+                     shorted ? node_index : node_capacitances_.size());
+        }
       }
-      double cap = parasitics_->nodeGndCap(node) + pinCapacitance(node);
-      node_capacitances_.push_back(cap);
     }
   }
 
+  // Lump each coupling capacitor to ground at any endpoint in the system.
```
parallaxsw: well no, that does not work with coupling capacitors.
at least respect the !parasitics_->isExternal(node1)) in the original code

dcalc/PrimaDelayCalc.cc
dsengupta0628: Thanks, but this actually had two call sites - the driver-seed loop and the resistor walk hence I had factored the shared placement logic out. I've now converted it from a capture-all lambda into a private member, placeNode(node, index,  queue), and cleaned up the surrounding function (dropped the second add_cap lambda in favor of the explicit node1/node2 form matching the original code). Will upload soon

dcalc/PrimaDelayCalc.cc
dsengupta0628: Okay fixed

dcalc/PrimaDelayCalc.cc
dsengupta0628: Binding node vars

dcalc/PrimaDelayCalc.cc
```diff
@@ -468,35 +468,79 @@ PrimaDelayCalc::findNodeCount()
...
-      size_t node_idx = node_index_map_.size();
-      node_index_map_[node] = node_idx;
-      const Pin *pin = parasitics_->pin(node);
-      if (pin) {
-        pin_node_map_[pin] = node_idx;
-        debugPrint(debug_, "ccs_dcalc", 1, "pin {} node {}",
-                   network_->pathName(pin), node_idx);
+  // Collect the nodes that enter G by walking out from the drivers through
+  // resistors. G is conductance-only, so a node with no resistive path to a
+  // driver has an all-zero row (singular) and is dropped.
```
jjcherry56: rows are not singular. matrices are.

no resistive path from a
driver has an all-zero row which is dropped to prevent singularity.

dcalc/PrimaDelayCalc.cc
```diff
@@ -468,35 +468,79 @@ PrimaDelayCalc::findNodeCount()
...
+  ParasiticNodeResistorMap resistor_map =
+      parasitics_->parasiticNodeResistorMap(parasitic_network_);
+  std::vector<ParasiticNode *> queue;
+  for (size_t drvr_idx = 0; drvr_idx < drvr_count_; drvr_idx++) {
+    const Pin *drvr_pin = (*dcalc_args_)[drvr_idx].drvrPin();
+    ParasiticNode *drvr_node =
+        parasitics_->findParasiticNode(parasitic_network_, drvr_pin);
+    if (drvr_node && !parasitics_->isExternal(drvr_node)
+        && node_index_map_.find(drvr_node) == node_index_map_.end())
```
jjcherry56: !contains

dcalc/PrimaDelayCalc.cc
```diff
@@ -468,35 +468,79 @@ PrimaDelayCalc::findNodeCount()
...
+    ParasiticNode *node = queue.back();
+    queue.pop_back();
+    size_t node_index = node_index_map_[node];
+    auto resistor_itr = resistor_map.find(node);
+    if (resistor_itr != resistor_map.end()) {
+      for (ParasiticResistor *resistor : resistor_itr->second) {
+        ParasiticNode *next_node = parasitics_->otherNode(resistor, node);
+        if (next_node
+            && !parasitics_->isExternal(next_node)
+            && node_index_map_.find(next_node) == node_index_map_.end()) {
```
jjcherry56: !contains

dcalc/PrimaDelayCalc.cc
```diff
@@ -569,13 +613,17 @@ PrimaDelayCalc::stampEqns()
...
-    ParasiticNode *node1 = parasitics_->node1(resistor);
-    ParasiticNode *node2 = parasitics_->node2(resistor);
-    // One commercial extractor creates resistors with identical from/to nodes.
-    if (node1 != node2) {
-      size_t node_idx1 = node_index_map_[node1];
-      size_t node_idx2 = node_index_map_[node2];
-      float resistance = parasitics_->value(resistor);
+    auto itr1 = node_index_map_.find(parasitics_->node1(resistor));
+    auto itr2 = node_index_map_.find(parasitics_->node2(resistor));
+    // Skip a resistor with an endpoint left out of the system.
```
jjcherry56: endpoint -> node
system -> network

PR #477, opened 2026-07-18
Fix/Avoid null deref in ArrivalVisitor::pruneCrprArrivals()

dsengupta0628:
pruneCrprArrivals() guards on hasCrprClkPin(), which only checks the crpr clk path handle is set - not that it still resolves to a live path. After incremental re-timing (e.g.  repair_design in OpenROAD) rebuilds clock vertices, the handle can go stale: hasCrprClkPin() stays true while crprClkPath() returns null, and when dereferenced with ->Path::minMax() we get segfaults.

Fix: resolve both crpr clk paths once into clk_path_no_crpr / clk_path_crpr and null-check both before calling minMax().

Caveat: crprClkPath() must never refer to a path that is not live - the null-check here is defensive against the stale handle, but the underlying invariant is that the handle should track only live paths.


jjcherry56:
I did something along these lines in
dc5ccd2d6 ArrivalVisitor::pruneCrprArrivals() seg fault (replace pr477)

It is a bit more direct about the min/max checking.

PR #478, opened 2026-07-23
MakeTimingModel: batch input arrival propagations

oharboe:
Batch input arrival propagation in MakeTimingModel::findTimingFromInputs.

Instead of running a separate search for every input pin and transition, assign virtual arrival clocks to batches of up to 64 inputs and run one propagation per batch. This significantly reduces timing graph traversals during model generation.


precisionmoon:
@dsengupta0628, could you review the code changes?  Thanks.

search/MakeTimingModel.cc
```diff
@@ -241,20 +242,24 @@ MakeTimingModel::checkClock(Clock *clk)
 class MakeEndTimingArcs : public PathEndVisitor
 {
 public:
-  MakeEndTimingArcs(Sta *sta);
+  MakeEndTimingArcs(const ModelClockSources &clk_sources,
+                    std::map<const Pin*, ClockEdgeDelays> &margins,
```
jjcherry56: the map should be typedef'd (using)

search/MakeTimingModel.cc
```diff
@@ -375,14 +434,20 @@ MakeTimingModel::findOutputDelays(const RiseFall *input_rf,
...
-          OutputDelays &delays = output_pin_delays[output_pin];
-          delays.delays.mergeValue(output_rf, min_max,
-                                   delayAsFloat(delay, min_max, sta_));
-          delays.rf_path_exists[input_rf->index()][output_rf->index()] = true;
+        const Clock *src_clk = path->clock(sta_);
+        if (src_clk) {
+          const auto src_itr = clk_sources.find(src_clk);
+          if (src_itr != clk_sources.end()) {
+            const Pin *input_pin = src_itr->second.first;
+            const RiseFall *input_rf = src_itr->second.second;
```
jjcherry56: use structured binding
auto [input_pin, input_rf] = src_itr->second;

search/MakeTimingModelPvt.hh
```diff
@@ -55,6 +57,10 @@ public:
 
 using ClockEdgeDelays = std::map<const ClockEdge*, RiseFallMinMax>;
 using OutputPinDelays = std::map<const Pin *, OutputDelays>;
+// Virtual clock -> (input pin, input transition) it stands for during
+// a batched input timing search.
```
jjcherry56, line 63: I can't understand what this is saying. I suspect the "it stands for" just confuses the sentence.

search/MakeTimingModelPvt.hh
```diff
@@ -74,15 +80,18 @@ private:
   void makePorts();
   void checkClock(Clock *clk);
   void findTimingFromInputs();
-  void findTimingFromInput(Port *input_port);
+  void findTimingFromInputsBatch(const std::vector<const Pin*> &batch,
```
jjcherry56: PinSeq

search/MakeTimingModelPvt.hh
```diff
@@ -74,15 +80,18 @@ private:
...
+                                 const std::vector<Clock*> &clks);
```
jjcherry56: ClockSeq

search/MakeTimingModelPvt.hh
```diff
@@ -74,15 +80,18 @@ private:
...
   void findClkedOutputPaths();
   void findClkTreeDelays();
   void makeClkTreePaths(LibertyPort *lib_port,
                         const MinMax *min_max,
                         TimingSense sense,
                         const ClkDelays &delays);
-  void findOutputDelays(const RiseFall *input_rf,
-                        OutputPinDelays &output_pin_delays);
+  void findOutputDelays(const ModelClockSources &clk_sources,
+                        std::map<const Pin*, OutputPinDelays> &output_delays);
```
jjcherry56: typedef (using)

jjcherry56:
Can you give an example of the run time with/without this change?

search/MakeTimingModelPvt.hh
oharboe: fixed

search/MakeTimingModelPvt.hh
oharboe: fixed

search/MakeTimingModelPvt.hh
oharboe: fixed

oharboe:
I discovered this while working on a design where this fix dropped write_timing_model from >2 hours down to a few minutes.

Synthetic Nangate45 benchmark results:
- 512 inputs: 1.27s -> 0.68s (1.9x)
- 2048 inputs: 9.20s -> 3.20s (2.9x)
- 8192 inputs: 68.90s -> 16.60s (4.2x)
- Deep dense cone (1024 inputs): 242.90s -> 82.10s (3.0x)

search/MakeTimingModelPvt.hh
oharboe: fixed

search/MakeTimingModel.cc
oharboe: fixed

search/MakeTimingModel.cc
oharboe: fixed

oharboe:
Better?

search/MakeTimingModel.cc
```diff
@@ -290,82 +293,138 @@ MakeEndTimingArcs::visit(PathEnd *path_end)
...
+        input_pins.push_back(input_pin);
+    }
   }
   delete port_iter;
+
+  if (input_pins.empty())
+    return;
+
+  // Path count per vertex scales with the batch size, so the batch
+  // size bounds search memory.
```
parallaxsw: huh??

search/MakeTimingModel.cc
```diff
@@ -290,82 +293,138 @@ MakeEndTimingArcs::visit(PathEnd *path_end)
...
+    + "_" + rf->shortName();
+  while (sdc_->findClock(name))
+    name += "$";
+  // Clone the default arrival clock waveform.
+  FloatSeq waveform;
+  waveform.push_back(0.0);
+  waveform.push_back(0.0);
+  sta_->makeClock(name, PinSet(network_), false, 0.0, waveform, "",
+                  sdc_->mode());
+  return sdc_->findClock(name);
```
parallaxsw: change Sta::makeClock return the clock instead of usiing findClock

search/MakeTimingModel.cc
oharboe: I will ask my AI to better itself and come up with a something coherent.

oharboe: Reworded in 74b13bb364bd: "Each vertex has an arrival per tag and each batched input adds a virtual clock tag, so cap the batch size to bound search memory."

search/MakeTimingModel.cc
oharboe: Fixed in 74b13bb364bd: Sta::makeClock now returns the Clock and makeBatchClock uses the return value.

search/MakeTimingModel.cc
```diff
@@ -290,82 +293,137 @@ MakeEndTimingArcs::visit(PathEnd *path_end)
...
+    + "_" + rf->shortName();
+  while (sdc_->findClock(name))
+    name += "$";
+  // Clone the default arrival clock waveform.
+  FloatSeq waveform;
+  waveform.push_back(0.0);
+  waveform.push_back(0.0);
+  return sta_->makeClock(name, PinSet(network_), false, 0.0, waveform, "",
+                         sdc_->mode());
 }
```
parallaxsw, line 386: std::string name = "make_timing_model_" + std::to_string(index)
    + "_" + rf->shortName();
  // Make sure group clock name is unique.
  while (sdc_->findClock(name))
    name += "_";
  // Clone the default arrival clock waveform.
  FloatSeq waveform{0.0, 0.0};
  PinSet pins;
  return sta_->makeClock(name, pins, false, 0.0, waveform, "", sdc_->mode());

oharboe: fixed.

oharboe:
@parallaxsw merged with latest master

jjcherry56:
I measured this on two designs and it is slower in both cases.
I tried changing the magic number for the batch size from 64 to 32 also.
Results in microseconds

jpeg asap7 20 inputs
hd 2691635
32 2677632
64 2719566

ascenium2 167M verilog 371 inputs
hd 19722241
32 32015364
64 30935411

oharboe:
> I measured this on two designs and it is slower in both cases. I tried changing the magic number for the batch size from 64 to 32 also. Results in microseconds
> 
> jpeg asap7 20 inputs hd 2691635 32 2677632 64 2719566

That's a pretty fast test case to begin with, less than a minute, I need to dig up the pathological case.

oharboe:
Dug up the pathological shape. The driver is the **product of input and output port count**, not cell count: `findTimingFromInputs()` runs one filtered arrival search per input port per transition, and each of those searches scans every output pin. Both of your designs (20 inputs, 371 inputs) sit far below the point where that product starts to hurt, which is why they show the batch overhead and none of the benefit.

Here is a self-contained case so you don't have to take my word for it — one Tcl file, no data, it generates its own netlist and reads `examples/nangate45_typ.lib.gz` out of this repo:

```tcl
# write_timing_model scaling reproducer (self-contained, Nangate45).
#
# Generates a wide, sparse-cone block: P input ports, each feeding its own
# short buffer column into its own flop; OUT output ports are taken off the
# first OUT flops.  This is the shape of a wide interface/channel block --
# very many input ports, a small logic cone per input, one large clock net.
#
# Usage: sta -no_init -exit repro.tcl
#   INPUTS=<n>  input ports           (default 1024)
#   OUTPUTS=<n> output ports          (default 8)
#   DEPTH=<n>   buffers per column    (default 4)
#   LIB=<path>  liberty               (default <repo>/examples/nangate45_typ.lib.gz)
#   WORK=<dir>  scratch dir           (default ./work)

proc opt {name def} {
  if {[info exists ::env($name)]} { return $::env($name) }
  return $def
}

set P     [opt INPUTS  1024]
set NOUT  [opt OUTPUTS 8]
set DEPTH [opt DEPTH   4]
set LIB   [opt LIB     [file join [file dirname [info script]] .. sta examples nangate45_typ.lib.gz]]
set WORK  [opt WORK    [file join [pwd] work]]

file mkdir $WORK
set tag ${P}_${NOUT}_${DEPTH}
set vfile [file join $WORK fabric_$tag.v]
set sfile [file join $WORK fabric_$tag.sdc]
set lfile [file join $WORK fabric_$tag.lib]

# ---- generate the netlist -------------------------------------------------
set t0 [clock milliseconds]
set f [open $vfile w]
puts $f "module fabric (clk, in, out);"
puts $f "  input clk;"
puts $f "  input \[[expr {$P-1}]:0\] in;"
puts $f "  output \[[expr {$NOUT-1}]:0\] out;"
puts $f "  wire \[[expr {$P*$DEPTH-1}]:0\] w;"
puts $f "  wire \[[expr {$P-1}]:0\] q;"
for {set i 0} {$i < $P} {incr i} {
  set src "in\[$i\]"
  for {set j 0} {$j < $DEPTH} {incr j} {
    set w "w\[[expr {$i*$DEPTH+$j}]\]"
    puts $f "  BUF_X1 b_${i}_${j} (.A($src), .Z($w));"
    set src $w
  }
  puts $f "  DFF_X1 r_$i (.D($src), .CK(clk), .Q(q\[$i\]));"
}
for {set i 0} {$i < $NOUT} {incr i} {
  puts $f "  BUF_X1 o_$i (.A(q\[[expr {$i % $P}]\]), .Z(out\[$i\]));"
}
puts $f "endmodule"
close $f

set f [open $sfile w]
puts $f "create_clock -name clk -period 500 \[get_ports clk\]"
close $f
set t_gen [expr {[clock milliseconds] - $t0}]

# ---- read + link ----------------------------------------------------------
set t0 [clock milliseconds]
read_liberty $LIB
read_verilog $vfile
link_design fabric
read_sdc $sfile
set t_read [expr {[clock milliseconds] - $t0}]

# ---- the measurement ------------------------------------------------------
set t0 [clock milliseconds]
write_timing_model $lfile
set t_model [expr {[clock milliseconds] - $t0}]

puts "RESULT inputs=$P outputs=$NOUT depth=$DEPTH cells=[expr {$P*($DEPTH+1)+$NOUT}] gen_ms=$t_gen read_ms=$t_read write_timing_model_ms=$t_model lib=$lfile"
```

Run one point with `INPUTS=8192 OUTPUTS=32768 sta -no_init -exit repro.tcl`.

The block shape is a wide interface/channel block: each input drives its own short buffer column into its own flop, so the per-input cone is tiny and the cones are disjoint; outputs come off the flops; one clock.

Measured on one host, single thread, `DEPTH=2`, master vs this PR:

| inputs | outputs | master | this PR | ratio |
| -----: | ------: | -----: | ------: | ----: |
| 20 | 8 | 1 ms | 2 ms | 0.5x |
| 371 | 371 | 43 ms | 45 ms | 0.96x |
| 1024 | 1024 | 189 ms | 200 ms | 0.95x |
| 2048 | 8192 | 1.54 s | 0.70 s | 2.2x |
| 8192 | 8192 | 5.98 s | 3.91 s | 1.5x |
| 8192 | 32768 | 41.6 s | 6.9 s | 6.0x |
| 16384 | 65536 | 229.8 s | 24.7 s | 9.3x |

The emitted `.lib` is byte-identical between the two binaries at every point above (76 MB at the largest), so this is purely a runtime difference.

So I think your measurements and mine are two ends of the same curve rather than a contradiction: below roughly a thousand ports the batch setup isn't repaid, and above it the amortized output-pin scan takes over.

I'm wary of adding conditional code on smaller port-counts, more iffy code is harder to maintain and test and on smaller batch sizes this doesn't make much of a difference anyway.

oharboe:
A real-world data point for where `write_timing_model` time goes.

XiangShan (Kunminghu-V3, the OpenXiangShan core) built hierarchically as 24 blocks on ASAP7 with OpenROAD c7e5e75 and its bundled OpenSTA, without this PR. Each block's abstract is `estimate_parasitics -placement` followed by `write_timing_model`, and the model dominates: 1 to 21 minutes per block, up to 30x the parasitics step. The blocks with the most pins take longest, not the blocks with the most cells:

| block | input pins | output pins | instances | estimate_parasitics | write_timing_model |
|---|---|---|---|---|---|
| LsqWrapper (load/store queues) | 3,877 | 3,692 | 1,099,661 | 78 s | 1,270 s |
| Rob | 2,150 | 1,062 | 1,017,003 | 91 s | 850 s |
| Dispatch | 3,453 | 8,204 | 308,775 | 22 s | 639 s |
| ExuBlock | 3,546 | 3,170 | 343,318 | 38 s | 627 s |
| DCacheWrapper | 3,265 | 2,503 | 560,981 | 60 s | 522 s |
| Bpu | 1,576 | 1,250 | 879,385 | 54 s | 357 s |
| Rename | 2,751 | 2,902 | 218,200 | 25 s | 298 s |
| VecRegionModule | 4,438 | 4,815 | 1,305,401 | 105 s | 194 s |
| Region_1 (fp) | 3,657 | 1,442 | 407,638 | 50 s | 151 s |
| twelve issue queues, each | 700 to 850 | 100 to 470 | 22k to 47k | | 53 to 119 s |

Dispatch at 309k instances takes three times longer than VecRegionModule at 1.3M, on 11,700 pins against 9,300; the vector region's outputs are mostly registered, so its input cones are shallow. That is the pins-times-cone-depth shape rather than a cell-count one, on a core that gets taped out. In a hierarchical flow it is the tail of every iteration: the parent cannot start until the last block's model is written, and by then the rest of the machine is idle.

PR #479, opened 2026-07-23
Extend user-defined properties to all basic object types

dsengupta0628:
### Summary

Follow-up to https://github.com/parallaxsw/OpenSTA/pull/456, which added define_property/set_property/get_property for MCMM types (Scene, Mode) only.

This extends user-defined properties to 10 more types: library, liberty_library, cell, liberty_cell, port, liberty_port, instance, pin, net, clock.

  **Changes**
  - search/Property.cc: explicit defineProperty<TYPE> template instantiations for the 10 types; new isUserProperty() helper.
  - search/Property.i: extend define_property_cmd and set_property_cmd dispatch to the new types.
  - getProperty() for each type: a declared-but-unset user property now reads back as an empty (none) value instead of throwing, matching Scene/Mode semantics. Gated on isUserProperty() so undefined properties still raise PropertyUnknown and
  externally-registered handlers are unaffected.
  - tcl/Property.tcl: update define_property -object_type help string.

New test/user_properties.tcl regression: set/get on pin, net, instance, clock; unset returns empty; get_* -filter skips unset objects without error; undefined property still errors.

  **Notes**
  - Behavior for existing types and undefined properties is unchanged; no default values are auto-registered.
  - Pre-existing quirk unchanged: get_property -object_type cell resolves via get_cells (instance), while define_property -object_type cell targets Cell.


include/sta/Property.hh
```diff
@@ -284,6 +284,11 @@ protected:
   PropertyValue::Type propertyType(std::string_view type);
   PropertyValue coercePropertyValue(PropertyValue::Type type,
                                     std::string_view value);
+  // True if a user-defined property of this name was declared (via
+  // defineProperty) on this object type, even if never set on a
+  // particular object.
```
jjcherry56: I think "even if never set on a particular object." is superfluous

search/Property.cc
```diff
@@ -1420,6 +1440,14 @@ Properties::coercePropertyValue(PropertyValue::Type type,
   }
 }
 
+bool
+Properties::isUserProperty(std::string_view object_type,
+                           std::string_view property)
+{
+  return prop_types_.find({std::string(object_type), std::string(property)})
+    != prop_types_.end();
```
jjcherry56: contains. finally an alternative to this blecherous c++ idiom

dsengupta0628:
Thanks for the review. I addressed them now

PR #482, opened 2026-08-02
Support gzipped SDC via libz (MOH-91)

akashlevy:
## Summary
- Fix `read_sdc` / `include` for `.sdc.gz` when Tcl has no `zlib` command (packaged Tcl 8.6 used by silisizer), matching liberty/verilog libz decompression ([MOH-91](https://linear.app/silimate/issue/MOH-91)).
- Add `ungzip_file` SWIG helper and `read_sdc_gzip` regression that renames `zlib` away to verify the libz path.
- Include `fmt/format.h` instead of `fmt/core.h` for fmt 12+ builds.

## Test plan
- [x] `./test/regression read_sdc_gzip` passes
- [ ] Confirm `read_sdc foo.sdc.gz` works in a silisizer build with Tcl that lacks `zlib`
- [ ] Sanity-check plain (non-gzip) `read_sdc` still works


akashlevy:
Closing: this change should land in Silimate/OpenSTA first, not upstream.

PR #484, opened 2026-08-04
Chunk vertices in BFS parallel iteration for dynamic load balancing

calewis:
Chunk vertices into fixed task-sized blocks (8 vertices per task) for dynamic load balancing across worker threads.

Performance Profiling Results (report_checks time on gcd_sky130hd):
- 1 Thread: 4.3% speedup (9458.35 us -> 9046.75 us)
- 2 Threads: 1.5% speedup (9442.85 us -> 9297.65 us)
- 4 Threads: Baseline performance (9402.20 us vs 9425.70 us)
- 8 Threads: 0.6% speedup (9337.25 us -> 9280.30 us)
- 16 Threads: 1.2% speedup (9870.90 us -> 9748.55 us)
- 32 Threads: 6.5% speedup (10067.80 us -> 9418.25 us)


calewis:
It always felt a bit odd to me that adding more threads would lead to skipping parallelism at larger and larger levels, the pathological case being that I have something like 48 threads and every level is exactly 47 vertices.   The original goal was to fix that, but this has the added bonus that it improve performance on the big levels as well, by more effective load balancing.  

I did some scans of an internal design to decide that 8 vertices was a good number.  The space of good chunk sizes looks like a U and 8 is close to the left side of the U while still being near the floor.  

Finally, we skip parallelism if we have fewer than 3 chunks mostly because the dispatch queue leaves the main thread sleeping so we want at least 3-4 (4 was slightly worse than 3 and 2 was worse than both) to wake up to do work. 

  > [!NOTE]
I think this a better way to distribute the tasks regardless of performance gains (that's just a bonus) so I'm asking you to take it for that reason, not because it's worth a few seconds in gcd_sky130hd.

calewis:
Ping @jjcherry56

parallaxsw:
This is problemated because I have 2 unmerged branches in development that change the BFS search. I do not want to spend my time resolving merge conflicts for trivial speedups.

calewis:
Do you have a link to the branches? Maybe it integrates cleanly?

parallaxsw:
I am not ready to make them public and will not any time soon.

PR #489, opened 2026-08-14
Verilog reader/link: fix crashes, a hang, report errors

dsengupta0628:
### Summary

 Six legal Verilog-2005 inputs  either killed  the process, spun forever, 
 or reported something a user can't act on. Behavior  before each fix was 
 measured on master, not inferred. 
Fixes one category of problems in https://github.com/The-OpenROAD-Project/OpenSTA/issues/399

  | # | Input | Before | After |
  |---|---|---|---|
  | 1 | `module top ({a}, z);` — concatenation in a module header | SIGSEGV in `link_design` | `Warning 2721` + `Warning 2722`, link fails cleanly |
  | 2 | `module sub (.px(m), .py(r));` instantiated positionally | SIGSEGV — pin vector resized to 0, then written at `pins_[SIZE_MAX]` | `Warning 2723`, link fails cleanly |
  | 3 | `assign q = p; assign p = q;` | `link_design` never returns | links correctly, `p`/`q` aliased |
  | 4 | `.A2('b0)` — unsized constant | `Error: stol`, no id/file/line | `Error: 2724 file line 8, unsized constant 'b0 not supported.` |
  | 5 | `find_pin` on a bundle port | returned a garbage pin handle, SIGSEGV on use | returns `NULL` |
  | 6 | `connect_pin a[0] u/b` on a bus port | no error, exit 0 — the instance silently lost every pin, and `write_verilog` then dropped a connection | `Error: 2728 port b has no pin to connect.` |

  ### Changes

  `verilog/VerilogReader.cc`
  - `makeCellPorts`: warn (2721) when a header entry has no port identifier
  - `linkNetwork`: null-check the port before `findPin`, link error 2722
  - `makeOrderedInstPins`: reject a bundle port, link error 2723
  - `makeInstPin`: don't hand a null pin to `makeTerm`
  - `parseConstant`: parse the constant width with `stringLong` instead of an
    unguarded `std::stol` — empty width is 2724, anything else unusable is 2725

  `network/ConcreteNetwork.cc`
  - `ConcreteNet::mergeInto`: return when `net == this`. The self-merge moved the
    net's pins and terms onto itself and left `merged_into_` pointing at `this`,
    which is what made walking the merge chain endless
  - `ConcreteInstance::findPin`: the name overload delegates to the port overload;
    both require a real pin index
  - `ConcreteNetwork::connect`: error 2728 on a port with no pin index, before a
    pin is created
  - `ConcreteInstance::addPin`: `criticalError` 2727 as an internal backstop
    rather than the out-of-bounds store

  `include/sta/ConcreteLibrary.hh`
  - `ConcretePort::hasPinIndex()`, so "bus/bundle ports have no pin index" is
    stated once rather than re-derived at each call site

  ### Test

  `test/verilog_link_errors.{tcl,ok}` with six `.v` files, one per case,
  registered in `test/regression_vars.tcl`. Full regression 53/53; no existing
  golden moved.

  New message ids: 2721-2725, 2727, 2728.


network/ConcreteNetwork.cc
```diff
@@ -1423,6 +1424,9 @@ ConcreteNetwork::connect(Instance *inst,
       disconnectNetPin(prev_net, cpin);
   }
   else {
+    // A bus or bundle port has no pin index, so it has no pin to connect.
+    if (!cport->hasPinIndex())
+      report()->error(2728, "port {} has no pin to connect.", cport->name());
```
parallaxsw: change to warn and skip making the pin

network/ConcreteNetwork.cc
```diff
@@ -1858,6 +1860,11 @@ ConcreteNet::ConcreteNet(std::string_view name,
 void
 ConcreteNet::mergeInto(ConcreteNet *net)
 {
+  // Merging a net into itself moves its pins and terms onto itself, nulls
+  // them, and leaves merged_into_ pointing at this, so following the merge
+  // chain never ends.
+  if (net == this)
+    return;
```
parallaxsw: make body conditional, with no return

verilog/VerilogReader.cc
```diff
@@ -1169,7 +1176,19 @@ VerilogNetConstant::parseConstant(std::string_view constant,
   std::string csize(constant.substr(0, csize_end));
 
   // Read the constant size.
-  size_t size = std::stol(csize);
+  size_t size = 1;
+  if (csize.empty())
+    // IEEE 1364-2005 3.5.1 makes the size optional; the width comes from the
+    // context, which is not known here.
+    reader->error(2724, reader->filename(), line,
```
parallaxsw: warn

verilog/VerilogReader.cc
```diff
@@ -1468,6 +1487,12 @@ VerilogReader::linkNetwork(std::string_view top_cell_name,
         while (net_name_iter->hasNext()) {
           const std::string &net_name = net_name_iter->next();
           Port *port = network_->findPort(top_cell, net_name);
+          if (port == nullptr) {
+            // No port was made for this module port list entry.
+            linkError(2722, module->filename(), module->line(),
```
parallaxsw: warn

verilog/VerilogReader.cc
```diff
@@ -1169,7 +1176,19 @@ VerilogNetConstant::parseConstant(std::string_view constant,
...
+  size_t size = 1;
+  if (csize.empty())
+    // IEEE 1364-2005 3.5.1 makes the size optional; the width comes from the
+    // context, which is not known here.
+    reader->error(2724, reader->filename(), line,
+                  "unsized constant {} not supported.", constant);
+  else {
+    auto [size_value, valid] = stringLong(csize);
+    if (!valid || size_value <= 0)
+      reader->error(2725, reader->filename(), line,
```
parallaxsw: warn

verilog/VerilogReader.cc
```diff
@@ -1657,6 +1682,15 @@ VerilogReader::makeOrderedInstPins(Cell *cell,
...
     VerilogNet *net = *pin_iter++;
     Port *port = port_iter->next();
+    // A bundle port from an explicit named module port (IEEE 1364-2005 A.1.4)
+    // has no pin index, so it cannot be connected.
+    if (network_->isBundle(port)) {
+      std::string inst_vname = instanceVerilogName(mod_inst->instanceName());
+      linkError(2723, parent_module->filename(), mod_inst->line(),
+                "instance {} port {} port expression cannot be connected.",
+                inst_vname, network_->name(port));
+      continue;
```
parallaxsw: don't use continue like this

parallaxsw:
I wget the .patch for PRs to test them locally.
Lately your PRs include a bunch of changes that are already merged that cause patch conflicts.
Please take a look at how you generate the PR so it doesn't include irrelevant changes.

network/ConcreteNetwork.cc
```diff
@@ -1423,6 +1424,9 @@ ConcreteNetwork::connect(Instance *inst,
       disconnectNetPin(prev_net, cpin);
   }
   else {
+    // A bus or bundle port has no pin index, so it has no pin to connect.
+    if (!cport->hasPinIndex())
```
parallaxsw: if (cport->isBus() || cport->isBundle())

remove ConcretPort::hasPinIndex - that is an artifact, not a property.

network/ConcreteNetwork.cc
```diff
@@ -1657,22 +1661,17 @@ ConcretePin *
...
 
 ConcretePin *
 ConcreteInstance::findPin(const Port *port) const
 {
   const ConcretePort *cport = reinterpret_cast<const ConcretePort*>(port);
-  size_t port_index = cport->pinIndex();
-  if (port_index < pins_.size())
-    return pins_[port_index];
+  if (cport
+      && cport->hasPinIndex()
```
parallaxsw: cport->isBus() || cport->isBundle()

network/ConcreteNetwork.cc
```diff
@@ -1758,8 +1757,11 @@ void
 ConcreteInstance::addPin(ConcretePin *pin)
 {
   ConcretePort *cport = reinterpret_cast<ConcretePort *>(pin->port());
-  size_t pin_index = cport->pinIndex();
-  if (pin_index >= pins_.size())
+  int pin_index = cport->pinIndex();
+  // Storing at pins_[-1] corrupts memory.
+  if (!cport->hasPinIndex())
```
parallaxsw: cport->isBus() || cport->isBundle()

test/verilog_link_errors.tcl
```diff
@@ -0,0 +1,24 @@
+# Verilog inputs that crashed, hung, or reported without a message id.
+#
+# Each .v says what it is for. Case 3 prints nothing; terminating is the
+# result being checked. Each case is read fresh because link_design deletes
+# the parsed modules. The catch is only so one case's error does not stop
+# the cases after it.
+
+read_liberty ../examples/nangate45_typ.lib.gz
+
+foreach { case cmds } {
```
parallaxsw: WTF?? just use separate tests. this is needlessly complicated

dsengupta0628:
> I wget the .patch for PRs to test them locally. Lately your PRs include a bunch of changes that are already merged that cause patch conflicts. Please take a look at how you generate the PR so it doesn't include irrelevant changes.

Thanks for flagging this - and I really apologize for this- you're right, and I found the cause.
 
I had been syncing my fork by merging parallaxsw/master into my local master rather than rebasing, and my master also carried unrelated work that had since landed upstream. GitHub therefore listed every one of those merge commits and the already-merged changes as part of this PR, so the generated .patch replayed hunks that are already in master. That's what 
 conflicts when you apply it. I've rebased the branch directly onto the current parallaxsw/master tip and squashed it to a single commit containing only the Verilog reader/link changes and their tests now. Going forward I'll rebase onto upstream instead of merging, so the PR only ever contains its own commits.

test/verilog_link_errors.tcl
dsengupta0628: Again- apologize! There were so many categories of the errors, that I thought your test directory will become very big- hence I bundled them into one for each category (from the various categories I plan to fix in upcoming weeks https://github.com/The-OpenROAD-Project/OpenSTA/issues/399- of course I will discuss with you offline first)- but I will simplify now and add each test of each failure.

doc/Messages.md
dsengupta0628, line 1: I had to use AI's help to fix a CI failure and seems it is causing CI to fail. So the doc/Messages.md is in the diff because the build seems to regenerate it and CI diffs it. Six lines: the two new ids (2727, 2728), plus four existing entries whose recorded file:line shifted  because my edits moved code - 1000, 1390, 1391, 1870. Not a hand edit.

  Worth noting separately: 2721-2725 don't appear in that table at all. etc/FindMessages.tcl matches the id and the format string on a single line, and those calls wrap across lines. Existing messages in VerilogReader.cc (197, 201, 202) are missing for the same reason, so it predates this PR and I left it alone

PR #490, opened 2026-08-16
improve quality of life for messages check

oharboe:
- Remove exact line numbers from generated Messages.md to prevent fragile CI failures when C++ files are modified.
- Enhance duplicate message ID detection to report the specific files and line numbers causing the collision.
- Enhance duplicate message ID detection to suggest the next available free ID for easy resolution.

I'm thinking the line numbers are not particularly valuable in the error messags as they could be different on origin/master and in a reported issue anyway: it can be trusted, but must be verified, so might as well leave out the line number from error messages.

### DIFF
--- a/etc/FindMessages.tcl
+++ b/etc/FindMessages.tcl
@@ -119,7 +119,7 @@ proc md_escape { text } {
   return [string map {| \\| ` \\`} $text]
 }
 
-proc report_msgs { {out_file ""} } {
+proc report_msgs { out_file include_lines } {
   global msgs
 
   if { $out_file != "" } {
@@ -136,7 +136,11 @@ proc report_msgs { {out_file ""} } {
   puts $out "| --- | --- | --- |"
   foreach msg_info $msgs {
     lassign $msg_info msg_id file line msg1
-    set loc "[file tail $file]"
+    if { $include_lines } {
+      set loc "[file tail $file]:$line"
+    } else {
+      set loc "[file tail $file]"
+    }
     puts $out "| [format %04d $msg_id] | $loc | [md_escape $msg1] |"
   }
   if { $out_file != "" } {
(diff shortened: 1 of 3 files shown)


oharboe:
@jjcherry56 If I understand correctly, on origin/master, any time you modify a C++ file you'd get an error because line numbers change without this fix, which I don't think is intentional.

Or am I on the wrong track here?

etc/FindMessages.tcl
```diff
@@ -81,14 +81,38 @@ proc check_msgs { } {
...
-      puts stderr "Warning: Message id $msg_id duplicated"
+      lappend duplicates $msg_id
       set has_error 1
     }
     set prev_id $msg_id
   }
+
+  set duplicates [lsort -unique -integer $duplicates]
+
+  if { [llength $duplicates] > 0 } {
```
parallaxsw, line 99: $messages != {}

etc/FindMessages.tcl
```diff
@@ -81,14 +81,38 @@ proc check_msgs { } {
...
+    foreach dup_id $duplicates {
+      puts stderr "Error: Message id $dup_id duplicated in:"
```
parallaxsw, line 101: It is not an error. It is a warning. See original code.

etc/FindMessages.tcl
```diff
@@ -112,7 +136,7 @@ proc report_msgs { {out_file ""} } {
   puts $out "| --- | --- | --- |"
   foreach msg_info $msgs {
     lassign $msg_info msg_id file line msg1
-    set loc "[file tail $file]:$line"
```
parallaxsw, line 115: do not remove the line numbers

parallaxsw:
I do not understand why you think there are errors or the build is "fragile" with line numbers.
The line numbers tell me where to find the error so they stay.

oharboe:
> I do not understand why you think there are errors or the build is "fragile" with line numbers. The line numbers tell me where to find the error so they stay.

I had to update messages.md with lots of line number changes unrelated to my PR.

oharboe:
I've updated the PR with a compromise that should satisfy both goals:

1. `doc/Messages.md` is still generated without line numbers. This keeps the git tree clean for contributors when they edit C++ files and prevents unrelated CI failures(or stale line numbers/unrelated line number changes to PR).
2. During the build, a new, untracked file `build/Messages-lines.txt` is generated that *does* contain the exact line numbers. 

This means that contributors don't have to deal with noisy diffs or fragile CI, but you (and any other maintainer/developer) can simply check your local `build/Messages-lines.txt` to immediately find the accurate source line for any error ID you encounter.

(I've also retained the helpful duplicate ID detection and suggestions). Does this look good to you?

oharboe:
Please reopen if this could be helpful

parallaxsw:
Messages.md is no longer in the repo which should address the issues you were having.

oharboe:
> Messages.md is no longer in the repo which should address the issues you were having.

Ah!  Makes sense. Thanks!

PR #491, opened 2026-08-17
FEATURE: Support Liberty retaining_rise/fall and retain slew tables

### LINKED ISSUE #310: Support for retaining_fall / retaining_rise
At the moment I'm a bit confused how to best describe behavior of a memory when for instance the data output becomes invalid after 1 ns but the new valid data output is only after 4 ns.

As far as I can tell the way to describe this is using `retaining_fall` / `retaining_rise` constraint, but AFAICT this is not supported at the moment.

Is there another way ? Or is this supported and I missed it ? Or is support planned ?

akashlevy:
## Summary
- Parse Liberty `retaining_rise` / `retaining_fall` delay tables and `retain_rise_slew` / `retain_fall_slew` slew tables in `timing()` groups.
- Use retain tables as min-path contamination delay (and retain slew when present); max delay still uses `cell_rise` / `cell_fall`.
- If retain slew is omitted, fall back to the parent delay-arc transition table. Matching LVF retain groups are parsed the same way as cell-delay LVF.

Fixes #310.

## Test plan
- [x] `./test/regression liberty_retain`
- [x] `./test/regression liberty_float_as_str liberty_latch3 liberty_arcs_one2one_1 delay_calc`
- [x] Confirm a commercial liberty with retain groups loads and min/max delays differ as expected


test/liberty_retain.tcl
```diff
@@ -0,0 +1,29 @@
...
+read_verilog liberty_retain.v
+link_design top
+
+create_clock -name clk -period 10
+set_input_delay -clock clk 0 {in1 in2}
+set_output_delay -clock clk 0 {out1 out2}
+set_input_transition 0 {in1 in2}
+set_load 0 {out1 out2}
+
+puts "max path"
```
parallaxsw: I don't see any point in these sentinels when the report has min/max in the header

liberty/TimingArc.cc
```diff
@@ -616,17 +643,38 @@ TimingArc::model(const Scene *scene,
...
       return scaled_model;
   }
   return scene_arc->model();
 }
 
+TimingModel *
+TimingArc::retainModel() const
+{
+  const RiseFall *to_rf = to_rf_->asRiseFall();
+  if (to_rf)
```
parallaxsw: why would to_rf_ be null?

liberty/LibertyReader.cc
```diff
@@ -2188,6 +2188,51 @@ LibertyReader::makeTableModels(LibertyCell *cell,
       found_model = true;
     }
 
+    TableModel *retain_delay_model =
```
parallaxsw: refactor this

include/sta/TimingArc.hh
```diff
@@ -119,6 +119,12 @@ public:
   TimingModel *model(const RiseFall *rf) const;
   void setModel(const RiseFall *rf,
                 TimingModel *model);
+  // retaining_rise/fall and retain_rise_slew/retain_fall_slew tables.
+  TimingModel *retainModel(const RiseFall *rf) const;
+  void setRetainModel(const RiseFall *rf,
+                      TimingModel *model);
+  // Cell/check model, or retain model if no cell/check model exists.
+  TimingModel *arcModel(const RiseFall *rf) const;
```
parallaxsw: what is the point of renaming "model" to "arcModel"?
It only makes sense if there is a non-arc model, and there isn't (that I know of)

include/sta/TimingArc.hh
```diff
@@ -119,6 +119,12 @@ public:
   TimingModel *model(const RiseFall *rf) const;
   void setModel(const RiseFall *rf,
                 TimingModel *model);
+  // retaining_rise/fall and retain_rise_slew/retain_fall_slew tables.
+  TimingModel *retainModel(const RiseFall *rf) const;
+  void setRetainModel(const RiseFall *rf,
+                      TimingModel *model);
+  // Cell/check model, or retain model if no cell/check model exists.
```
parallaxsw: It doesn't make sense to me arcModel would ever point to a retain model if no cell/check model exists.

include/sta/TimingArc.hh
akashlevy: Reverted. `LibertyBuilder` uses `model()` again.

liberty/LibertyReader.cc
akashlevy: Extracted retain table/LVF construction into `makeRetainTableModels()`.

test/liberty_retain.tcl
akashlevy: Dropped the max/min path `puts`; `Path Type` in the report header already distinguishes them. Left the dcalc labels because those reports do not include min/max in the header.

liberty/TimingArc.cc
akashlevy: Agreed — `to_rf_` is always set. Dropped the null check and look up `retainModel(to_rf_->asRiseFall())` directly.

include/sta/TimingArc.hh
akashlevy: Removed the fallback. Retain tables stay on `retainModel()` and are only used for min delay; they no longer substitute for a missing cell/check model.

akashlevy (commented):
@jjcherry56 look a bit better now?

PR #503, opened 2026-08-20
Use HTTPS for CentOS 7 vault yum repos

akashlevy:
## Summary
- The `linux-centos` job fails installing `devtoolset-11` because `http://vault.centos.org` now 301s to HTTPS, and yum on CentOS 7 cannot follow that redirect.
- Keep the existing vault paths and switch yum `baseurl`s to `https://vault.centos.org`.

## Test plan
- [ ] `linux-centos` builds `Dockerfile.centos7` and runs `test/regression`
- [ ] Ubuntu and macOS CI jobs still pass


akashlevy:
Seems the issue was intermittent, I'll close this for now.

PR #506, opened 2026-08-27
[ENG-2613] Liberty-based select power dump

stanminlee:
## Summary
- After `read_saif`, `report_power` still priced the whole netlist. Blackboxes / other Liberty libs with no useful power models made that walk expensive and noisy.
- Add `report_power -exclude_libs patterns`: report each leaf whose Liberty library name matches **none** of the glob patterns (same per-instance table / JSON path as `-instances`). Instances without a Liberty cell are skipped. If both `-instances` and `-exclude_libs` are given, `-instances` wins.
- Price and cache a single leaf in `Power::power()` instead of calling `ensureInstPowers()` (full netlist). Hierarchical instances still fill the cache. Clear `instance_powers_` / `instance_powers_valid_` on activity, scene, and `powerInvalid()` so a later `read_saif` or `set_power_activity` does not reuse stale leaf results.


stanminlee:
Sorry, accident

PR #514, opened 2026-09-14
Use the src attribute from pin in path reporting if it exists

mikesinouye:
To retain original RTL src annotations after MBFF clustering for each bit, report the src attribute if it exists on Pin, defaulting back to use Inst. Else, all bits within the MBFF report the same src attribute and cannot be differentiated.


jjcherry56:
this needs a regression test. test/verilog_attribute does not have any pin attributes (just ports).

mikesinouye:
Hi @jjcherry56 . Thanks for the attribute/property alignment commits. I have integrated those changes and re-implemented the feature, with a regression test using that work.

PR #516, opened 2026-09-17
Report the tcl traceback for an error while reading a file

dsengupta0628:
### Summary
include_file reads a file one complete command at a time and evaluates each with catch {uplevel #0 $cmd} result. 
On failure it keeps $result and re-raises it as Error: <file>, <line> <msg>. 
The errorInfo tcl had already built inside that catch is dropped, and because the reader evaluates command by command rather than handing the file to source, that catch is the only place it exists.

So an error a few proc frames down in an SDC or a command file reports as a single line that doesn't say which proc raised it. Recovering that by hand means re-running with puts probes, which is slow for a person and slower for automation driving a flow. The information is already computed; only the capture is missing.

What it looks like for a file that fails two proc frames down:

_**today:**_

`Error: constraints.sdc, 41 divide by zero`

_**with sta_error_traceback 1:**_
```
Error: constraints.sdc, 41 divide by zero
    while executing
"expr { $period / $ratio }"
    (procedure "derive_clock" line 3)
    invoked from within
"derive_clock core_clk"
```

**The change**

sta_error_traceback, 0 by default. 
When it is 1, catch is given its options dict and -errorinfo is used in place of the bare result.

A small helper trims the uplevel frames include_file itself contributes, so the traceback ends at the command read from the file.
read_sdc and include both route through this proc, so one change covers all three readers.

With the variable at 0 the reported text is unchanged.

Test added to demonstrate the enhancement.

**Notes**

- I defaulted it to 0 so nothing in existing output moves. FWIW, only three .ok files under test/ contain a line starting with Error, all sta_error-formatted, none in the file, or lines this proc emits, so defaulting to 1 would be close to free here. I left it off because downstream consumers may diff logs.
- For an error raised through sta_error, the traceback includes sta_error's own frames ahead of the useful ones. That could be trimmed as well, but it would make this larger than it needs to be for a first pass. Happy to follow up if you'd like it.
- trim_traceback matches the traceback text rather than anything structural. It takes the last occurrence of the marker, so a user's own uplevel frames are preserved and a failure at file top level trims cleanly. If you'd rather not have that coupling, dropping the helper and reporting -errorinfo unmodified is a one-line change; the cost is three extra lines at the end of every reported error.


dsengupta0628:
Two follow-ups on the same proc, as separate commits so they can be taken
or dropped independently of the traceback change.

**Start line of an incomplete command.** An unclosed brace gave
"incomplete command at end of file" with no location. The reader already
knows where the unfinished command began, so it now appends
"starting on line 5". Old wording kept, so existing matches still work.

**Start line of a multi-line command that failed.** include_line names a
command's *last* line, so an error inside an if opening on line 3 and
closing on line 7 was reported as line 7, and a warning from a command
continued across lines 14 to 18 as line 18. Both now report the line the
command starts on - in include_file's own message and in sdc_file_line,
so messages with a msg id get it too. Single line commands are
unaffected.

oharboe:
@dsengupta0628 Heads up! https://github.com/parallaxsw/OpenSTA/pull/517 was just merged, so you probably want to merge with master to see if there are any interactions. I don't have any specific reason to believe this, just a general precatuion because both PRs are concerned with normalizing Tcl behavior.

dsengupta0628:
> @dsengupta0628 Heads up! #517 was just merged, so you probably want to merge with master to see if there are any interactions. I don't have any specific reason to believe this, just a general precatuion because both PRs are concerned with normalizing Tcl behavior.

Thanks! There was no conflict with 517- I merged with latest master

PR #517, opened 2026-09-18
read_sdc scopes the .sdc unknown handler

oharboe:
An `.sdc` file is a domain specific language that is almost tcl. The one
reading it needs that tcl does not give is the bus subscript: in `foo[2]`
the brackets are part of a name, where tcl sees a command substitution.
`sta_unknown` supplies that reading. `tcl/Util.tcl` installed it as the
global namespace unknown handler when the file loaded, so every tcl
script a session runs was read that way, not only the `.sdc`.

`read_sdc` now installs the handler for the extent of the `.sdc` file and
puts the previous one back on the way out. It is restored when the `.sdc`
fails, and nested `read_sdc` calls restore in order. `include_file` is
still the reader and the file and line reporting behind the SDC message
ids is unchanged. Inside `read_sdc` every `.sdc` reads as before. Outside
`read_sdc` the interpreter is stock tcl.

## Where the line goes

OpenSTA has `.tcl` scripts of its own that have to change now, and
`pvt/test/` may have more. Whether OpenSTA should keep the tcl language
modifications outside an `.sdc` file is your call and I cannot judge it.
I chose the rule that is simplest to document: an `.sdc` file is a DSL,
and inside it the tcl language modifications are idiomatic. Outside an
`.sdc` file it is plain tcl.

## The leak was load bearing in OpenROAD

`write_pin_placement` wrote `place_pin -pin_name R0_addr[0]` with the
name unquoted, so the file the tool writes only read back on the leaked
interpreter. 27 tcl call sites named a command by a unique prefix of its
real name, such as `utl::metric_int` for `utl::metric_integer`. None of
it was deliberate, and none of it is visible until the interpreter is
stock tcl.

Those are fixed upstream in The-OpenROAD-Project/OpenROAD#11437 and
The-OpenROAD-Project/OpenROAD#11439 and
The-OpenROAD-Project/OpenROAD-flow-scripts#4542. OpenROAD tcl reads on a
stock interpreter now, so this change costs it nothing.

## Two commits

The first commit only changes tests. It spells out the four tests that
depend on the handler being global, so they read the same on a stock tcl
interpreter. It stands on its own and passes with the handler still
global.

The second commit scopes the handler.

## Tests

Four tests here depended on the handler being global. `get_is_memory`
called `link`, which is a unique abbreviation of `link_design`.
`disconnect_mcp_pin` used bare `data_in[1]` and `data_in[*]` names.
`liberty_arcs_one2one_1` and `liberty_arcs_one2one_2` built a pin name
as `A[$i]`, where the brackets are a command substitution and the
handler returned the subscript. All four are spelled out now. None of
the names appear in a command position in the logs, so the `.ok` files
are unchanged.

Those are all of them in `test/`. I cannot see
`pvt/test/`, so that tree needs a run. A test that depends on the
handler fails with `invalid command name` and the error names the file
and line.

## On regression coverage

No regression test is added for this.

If the handler is not installed inside `read_sdc`, a bare bus subscript
in an `.sdc` file fails with `invalid command name "0"` and the error
names the file and line. The fix is to brace the name. The
`get_is_memory` and `disconnect_mcp_pin` tests fail this way if the
handler is not installed.

OpenROAD reads its tcl scripts on this interpreter for every design in
its flows. It found the call sites that depended on the handler being
global. It will find a regression here, and the fix is one line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


oharboe:
@tspyrou @maliberty @dsengupta0628 Plug .sdc DSL leakage

maliberty:
@jjcherry56 I'm surprised you accepted this breaking change.  Any use of unescaped bus notation or command shortening will now fail.  I think this should be reverted and Google agrees.

gadfort:
This does seem to change the user behavior which should only really be done for a good reason and with a transition period / announcement.  Any scripts that were written before this may not work anymore without having to go in and fix them one by one.

parallaxsw:
It is done with a good reason or I would not have done it.
To the best of my knowledge It is the way commercial tools work.
I changed to warn that it will change in a future release instead of error in commit 7633e96f.

